PR #11, opened 2018-07-24
Windows compatibility

braybaud:
Hello,

I was trying to run on Windows and was getting an error.
The offending line is (https://github.com/nextstrain/cli/blob/master/nextstrain/cli/runner/docker.py#L105):
``` python
 # Run the process in the container with the same UID/GID so that file
 # ownership is correct in the bind mount directories.
 "--user=%d:%d" % (os.getuid(), os.getgid()),
```

Because `os.getuid()` and `os.getgid()` does not exist on Windows platforms. 

Opening a Pull Request to go around this limitation by only adding the `--user` argument if not on Windows. The file rights are much more lenient on Windows and therefore it is unnecessary to run as the current user. 

Also added some ignore files in the `.gitignore` for all the JetBrains generated files (useful if somebody is using PyCharm to develop).

Thanks,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,18 @@ __pycache__/
 Icon?
 ehthumbs.db
 Thumbs.db
+
+ Covers JetBrains PyCharm
+ ######################
+.idea/**/workspace.xml
+.idea/**/tasks.xml
+.idea/**/usage.statistics.xml
+.idea/**/dictionaries
+.idea/**/shelf
+.idea/**/dataSources/
+.idea/**/dataSources.ids
+.idea/**/dataSources.local.xml
+.idea/**/sqlDataSources.xml
+.idea/**/dynamic.xml
+.idea/**/uiDesigner.xml
+.idea/**/dbnavigator.xml
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -1,14 +1,13 @@
 """
 Run commands inside a container image using Docker.
 """
-
+import argparse
 import os
 import shutil
-import argparse
 import subprocess
 from collections import namedtuple
-from ..util import warn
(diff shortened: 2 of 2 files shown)


tsibley:
Thanks for the PR, @braybaud!  I'm going to apply the change you suggest manually.  The patch you provided also changes unrelated things like formatting, which is not great commit hygiene.

braybaud:
@tsibley Sounds good, thanks 👍

tsibley:
@braybaud Would you give the `windows` branch a try and report back?

I'm currently downloading a Windows environment to test in myself, but it'll be a little bit before that's ready still.  I also plan to setup automated testing on Windows so this doesn't happen in the future!  I appreciate you writing in with the bug report and a patch. :slightly_smiling_face: 

As for the `.gitignore` lines, I recommend adding those to your local `.git/info/exclude` file in your clone of this repo. The `.idea/` files are not specific to this repo, so I don't think they belong in the repo's `.gitignore`. If you want to always ignore those files in all of your repos, you can setup a global git ignore for yourself like this:

```
git config --global core.excludesfile %USERPROFILE%\.gitignore_global
```

Then edit the `%USERPROFILE%\.gitignore_global` file to add the patterns you want. They'll be ignored in all repos you clone.

braybaud:
Thank you I did not know about the this little git magic :)

I gave a try to the `windows` branch and was able to successfully build and view the Zika examples. 
Thanks again for your prompt reply!

tsibley:
Thank you for testing! I've released that fix as version 1.1.8 on PyPi. You'll be able to upgrade with pip with something like:

    pip install --upgrade nextstrain-cli

trvrb:
Thanks so much for the PR @braybaud. We really appreciate the help.

PR #12, opened 2018-07-24
view: Optionally allow remote network access to the webserver

tsibley:
This lets folks run `nextstrain view` on their computer and then share
the link with others on their network.  The original motivation for this
feature is to let Ashley run Nextstrain on her computer and have her
students connect to it from their computers.  However, it's also easy to
see how it would be useful in development when working collaboratively.

The default is still local access only so that the command is secure by
default.

This adds the first external dependency to the package.  Though the
feature and dep are non-essential, I opted not to make it part of an
optional feature tag under the reasoning that "batteries included" was a
better fit for our users.  If dependencies cause installation woes down
the road, then we should reconsider that choice.

### DIFF
--- a/nextstrain/cli/command/view.py
+++ b/nextstrain/cli/command/view.py
@@ -11,6 +11,7 @@ check-setup` to check if Docker is installed and works.
 """
 
 import re
+import netifaces as net
 from pathlib import Path
 from ..runner import docker
 from ..util import colored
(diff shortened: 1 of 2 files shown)


tsibley:
Resolves #12

huddlej:
@tsibley This works for @ashandrsn and me! Thanks for adding this feature so quickly!

PR #18, opened 2018-07-30
Check for newer versions of the CLI itself…

### LINKED ISSUE #16: Update command should check for new versions of CLI package

tsibley:
…in the check-setup and update commands.  These two commands already
come with the expectation that self-tests will be run and network
connections made, so it seems not very intrusive to do our own version
checking as well.

The idea is to prompt people to upgrade the CLI itself when they're
already thinking about their setup and not to distract or nag them when
they're running builds or viewing results.  Having people stay
relatively up-to-date on the CLI versions should make our development
easier.

Resolves #16.

### DIFF
--- a/nextstrain/cli/util.py
+++ b/nextstrain/cli/util.py
@@ -1,5 +1,9 @@
 import re
+import requests
+from pkg_resources import parse_version
 from sys import stderr
+from .__version__ import __version__
+
 
 def warn(*args):
     print(*args, file = stderr)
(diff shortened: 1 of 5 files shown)


tsibley:
@jameshadfield We chatted briefly about this idea this morning, and so I went about implementing it before forgetting. Your review would be appreciated.

nextstrain/cli/util.py
```diff
@@ -24,3 +28,33 @@ def colored(color, text):
...
+    newer_version = new_version_available()
+
+    if newer_version:
+        print("A new version of nextstrain-cli, %s, is available!  You're running %s." % (newer_version, __version__))
+        print()
+        print("Upgrade by running:")
+        print()
+        print("    pip install --upgrade nextstrain-cli")
+        print()
+
```
jameshadfield: Can you add an `else` statement to print "the nextstrain CLI is up-to-date" (or similar) here.

tsibley: Updated with your suggestion, as well as [some additional wording at the end of the `update` command](https://github.com/nextstrain/cli/pull/18/commits/818c1eca1a06905169e6ddd4d7622b34452ae8dc#diff-c6baea57f92567a217fe049316ce37a0R34).

tsibley:
@jameshadfield Further thoughts?

jameshadfield:
Looks fine. Can't review it for some reason.

tsibley:
FWIW, I think you can start a Review by going to the [Files changed](https://github.com/nextstrain/cli/pull/18/files) tab

PR #27, opened 2018-08-28
Native runner (and changes driven by it)

tsibley:
This PR adds support for `nextstrain build --native …` which runs the build outside of the container image.  Support for this is on my roadmap to AWS Batch builds, but it's also nice to just provide a consistent interface for running builds under different environments.

Several related refactorings and improvements to help messaging were driven by the `--native` addition (with an eye on supporting more runners soon), and those are included as separate commits bookending `--native`.  I recommend reviewing commit-by-commit from the start rather than the PR as a whole, as the story will be clearer.


trvrb:
This is fantastic @tsibley. Everything looks good to me. And working well for builds that I tested with. I like the interface.

Issue #28, opened 2018-09-11
Run builds on AWS Batch [issue]


tsibley:
This issue is for a work-in-progress feature which I've been working on recently and am currently polishing up.

Even with the working mechanics, there are several external things that need consideration before this can be considered a shippable feature:

- [x] CLI: Default region? (Region is required by `boto3.client("batch")`.)
- [x] IAM: policies allowing normal user roles to submit jobs
- [x] IAM: role/policies for Batch service (_is it limited enough?_)
- [x] IAM: role/policies for Batch jobs (_is it limited enough?_)
- [x] Compute environment: Put limits on resources used / costs incurred by Batch jobs to prevent runaways
- [x] Compute environment: Increase available CPU / memory resources
- [x] S3: Create build context bucket
- [x] S3: Add retention policy to S3 bucket
- [x] ~S3: Bucket ACLS - https://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html~
- [x] CloudWatch Logs: Add retention policy for Batch jobs
- [x] CLI: Delete log stream when done
- [x] Documentation: Creation of Batch compute environment, queue, job definition, S3 bucket, S3 retention policies, CloudWatch logs retention policies, and IAM roles/policies (+ automated tooling for doing this?)
- [x] Documentation: Describe security environment and assumptions
- [x] docker-base: Merge `aws-batch` branch
- [x] Batch: Switch job definition to use `nextstrain/base:latest` instead of `nextstrain/base:branch-aws-batch`
- [x] CLI: Terminate remote jobs on `^C` (or make issue for this)

(The list above is as much for me as anyone else.)

tsibley:
I've just pushed the CLI changes to adding support for `nextstrain build --aws-batch`: https://github.com/nextstrain/cli/compare/aws-batch.

I would appreciate any review of the code and especially the user interaction/experience.  Until I complete some of the ancillary items on the todo list above, it will only work if you have admin access to the lab's AWS account (this means @trvrb and maybe @jameshadfield for now).  I will leave another comment when I've arranged wider access so more folks can test if they want.

Note that the remote jobs are currently using the `nextstrain/base:branch-aws-batch` image from the `aws-branch` of `docker-base`: https://github.com/nextstrain/docker-base/compare/aws-batch.  Relevant entrypoint code is there.

tsibley:
[...]

PR #29, opened 2018-09-21
Run builds on AWS Batch [PR]

tsibley:
See my comments on #28 and the commit messages themselves.

### DIFF
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -178,7 +178,8 @@ S3 bucket name.
             "Effect": "Allow",
             "Action": [
                 "logs:GetLogEvents",
-                "logs:FilterLogEvents"
+                "logs:FilterLogEvents",
+                "logs:DeleteLogStream"
             ],
             "Resource": [
                 "arn:aws:logs:*:*:log-group:/aws/batch/job",
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -8,9 +8,9 @@ account is required, but only as an initial, one-time step.  See below for
 details.
 
 Launching Nextstrain builds on Batch from your computer is done using the
-`--aws-batch` flags to `nextstrain build`, for example:
+`--aws-batch` flag to `nextstrain build`, for example:
 
-    nextstrain build --aws-batch --aws-batch-s3-bucket=... zika-tutorial/
+    nextstrain build --aws-batch zika-tutorial/
 
 This uploads the [`zika-tutorial/` directory][] to S3, submits the
 Batch job, monitors the job status, streams the job logs to your terminal, and
@@ -65,15 +65,16 @@ to `export` the environment variable.
 The Nextstrain CLI's AWS Batch support must be told, at a minimum, the name of
 your S3 bucket (which you'll create below).
 
-You can do this by setting the `NEXTSTRAIN_AWS_BATCH_S3_BUCKET` environment
-variable
+You can do this by putting your bucket name in an environment variable
 
     export NEXTSTRAIN_AWS_BATCH_S3_BUCKET=...
 
-or passing the `--aws-batch-s3-bucket=...` option to `nextstrain build`.
+or in the `~/.nextstrain/config` file
 
-In the future, the CLI will support setting your bucket name in a persistent
-configuration file.
+    [aws-batch]
+    s3-bucket = ...
+
+or passing the `--aws-batch-s3-bucket=...` option to `nextstrain build`.
 
 
 # Setting up AWS to run Nextstrain builds
@@ -223,7 +224,8 @@ setup the job definition, compute environment, and job queue described below.
 
 Create a new job definition with the name `nextstrain-job`.  If you use a
 different name, you'll need to use the `--aws-batch-job` option to `nextstrain
-build` or set `NEXTSTRAIN_AWS_BATCH_JOB`.
+build`, set the `NEXTSTRAIN_AWS_BATCH_JOB` environment variable, or set `job`
+in the `[aws-batch]` section of `~/.nextstrain/config`.
 
 Choose the job role _NextstrainJobsRole_, which you just created in the IAM
 roles section above.
@@ -257,8 +259,9 @@ costs when no jobs are running.
 ### Job queue
 
 Create a job queue named `nextstrain-job-queue`.  If you use a different name,
-you'll need to use the `--aws-batch-queue` option to `nextstrain build` or set
-`NEXTSTRAIN_AWS_BATCH_QUEUE`.
+you'll need to use the `--aws-batch-queue` option to `nextstrain build`, set
+the `NEXTSTRAIN_AWS_BATCH_QUEUE` environment variable, or set `queue` in the
+`[aws-batch]` section of `~/.nextstrain/config`.
 
 If you're not using the wizard, make sure you connect the job queue to the
 compute environment you created above.
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -1,0 +1,305 @@
+# Running Nextstrain builds on AWS Batch
+
+The Nextstrain CLI supports launching pathogen builds on [AWS Batch][] from
+your own computer.  No local computational infrastructure is required when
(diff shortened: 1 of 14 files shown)


trvrb:
@tsibley ---

I'm trying to pick through this now (finally). One thing as I work:

* I'm sure this is in the commit messages somewhere, but I need a more obvious entry point. I figured out from issue #28 that I could run this via `nextstrain build --aws-batch .`, but I still don't know the additional options I can provide. I don't see anything about `--aws-batch` in `nextstrain build --help`. Is there some other place I should be looking?

trvrb:
I'm also getting the following error repeatedly, which I will dig into further but wanted to bring it up now:

```
Cleaning up logs
  File "/Users/trvrb/.pyenv/versions/3.6.5/bin/nextstrain", line 11, in <module>
    load_entry_point('nextstrain-cli==1.5.0', 'console_scripts', 'nextstrain')()
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/__main__.py", line 10, in main
    return cli.run( argv[1:] )
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/__init__.py", line 51, in run
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/command/build.py", line 67, in run
    return runner.run(opts, working_volume = opts.build)
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/runner/__init__.py", line 128, in run
    return opts.__runner__.run(opts, argv, working_volume = working_volume)
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/runner/aws_batch/__init__.py", line 184, in run
    logs.delete_stream(job.log_stream)
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/nextstrain/cli/runner/aws_batch/logs.py", line 46, in delete_stream
    logStreamName = stream)
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/trvrb/.pyenv/versions/3.6.5/lib/python3.6/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the DeleteLogStream operation: User: arn:aws:iam::827581582529:user/trvrb is not authorized to perform: logs:DeleteLogStream on resource: arn:aws:logs:us-east-1:827581582529:log-group:/aws/batch/job:log-stream:nextstrain-job/default/7e4b6628-4c87-4911-91f3-435fee389286
```

tsibley:
Thanks for digging in!

Hmm, [`nextstrain build --help` output](https://gist.github.com/ec41ef342f6e1c4a73e670cb853b3bc6) mentions AWS Batch support in several places. If it doesn't for you, then I think maybe you're not running the code on the branch?

AWS Batch support is also documented at <https://github.com/nextstrain/cli/blob/aws-batch/doc/aws-batch.md>. I mentioned this [in a comment on #28](https://github.com/nextstrain/cli/issues/28#issuecomment-423363499), and it's also referenced from the output of `nextstrain build --help-all`. For your review, it's definitely worth reading the other comments on #28. They're aimed at you, mostly. :-)

I believe the AccessDeniedException you were seeing is a result of the AWS IAM policy not including the DeleteLogStream action. I've fixed that and fixed the docs. Thanks for the catch! (I wasn't seeing them because I didn't test again with my non-privileged AWS user after adding the log stream cleanup.)

jameshadfield:
Running this with (what I think are) the default options doesn't get started for me :(

```
nextstrain build --aws-batch --aws-batch-job nextstrain-job --aws-batch-queue nextstrain-job-queue --aws-batch-s3-bucket nextstrain-job .

Nextstrain Run ID: a619b8d3-6b0c-4d8f-ab1c-c2ba760c386b
Uploading /Users/naboo/github/blab/zika-colombia to S3
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/bin/nextstrain", line 11, in <module>
    load_entry_point('nextstrain-cli', 'console_scripts', 'nextstrain')()
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/__main__.py", line 10, in main
    return cli.run( argv[1:] )
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/__init__.py", line 51, in run
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/command/build.py", line 67, in run
    return runner.run(opts, working_volume = opts.build)
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/runner/__init__.py", line 128, in run
    return opts.__runner__.run(opts, argv, working_volume = working_volume)
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/runner/aws_batch/__init__.py", line 72, in run
    bucket = s3.bucket(opts.s3_bucket)
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/runner/aws_batch/s3.py", line 159, in bucket
    raise ValueError('Bucket named "%s" does not exist' % bucket.name)
ValueError: Bucket named "nextstrain-job" does not exist
```

jameshadfield:
I've tested this functionality and it can report that it works for me.

I have a couple of suggestions:

(a) The documentation is good, however a flow diagram of the different components (cli, s3 bucket, batch, logs...), how they interact with each other, which permissions are used where etcetera would be extremely helpful (to someone like me). 

(b) Currently quitting the `nextstrain` program while running results in the job being cancelled and the data deleted from the bucket.
```
Job now RUNNABLE
^C
Canceling job…
Waiting for job to stop…
(Press Ctrl-C again if you don't want to wait.)
Job FAILED after 1.1 minutes (stopped by user)
Cleaning up S3
deleted: s3://nextstrain-jobs/ddb25a16-62cd-44eb-9021-3d3096971e23.zip
```
I imagine scenarios where one wants to stop the `nextstrain` process but keep the job running. Is this possible? How does this work if internet connection drops?

tsibley:
Thanks for testing this!  The feedback is very useful.

(a) Ah, yes, that'd be good. I've created #34 for this.

(b) I've previously sketched out ideas for supporting background jobs and disconnections, but they're not part of this initial implementation. The basic idea is an `--unattended` mode for AWS Batch builds with support for resuming/fetching results later using something like `nextstrain build --aws-batch --resume job-id …`.  This could be wired up automatically if a connection error happens.

(c) Orphaned data files and logs are left around if the local `nextstrain` client throws an unexpected exception. This is by design, to aid with debugging the client. Both the S3 bucket and CloudFront log group have retention lifecycles defined so that orphaned files/logs are removed automatically after 30 days if not proactively removed by the client first.

(d) It would be possible. When would this be useful?

(e) Log streams and job _ids_ cannot; their ids are assigned automatically. The _build_ id (or job _name_) and zip file on S3 could incorporate a username, however. Would that still be useful? I don't really expect folks to be accessing the AWS console in normal use cases.

jameshadfield:
I may not be the target audience for this tool and so my suggestions may be off track. I'd use this to submit jobs to aws in the "--unattended" mode, and perhaps dozens at a time. So I'd lose track of job IDs if I had to manually keep track of them. Something like `job-status` (which lists submitted jobs and their progress, including those that have finished) and `retrieve-finished` (which downloads the data) would be useful. Keeping around logs on AWS for ~30 days (and being able to subset by user) would help see what I'd run, similar to `job-status`.

tsibley:
Nod. What's currently there are the fundamental bits, but my intent with the `--unattended` mode would be to tackle your more specific use case, albeit incrementally. Saving job ids and using them automatically later is possible and definitely desirable, but I postponed it for now because there are some technical hurdles to doing it robustly esp. around fetching results of multiple runs for the same build.

PR #38, opened 2018-12-27
Run snakemake with '-p' flag to print commands

trvrb:
Running as `snakemake -p` causes snakemake to print shell commands run. I think this is generally useful to see what's going on and I run all local snakemake calls with `-p`.


tsibley:
:+1: I tweaked it slightly https://github.com/nextstrain/cli/pull/38/commits/21a07aa72bba35a1121cba441203f5cc602a7c36 and merged. Thanks!

PR #39, opened 2019-01-18
view: Update for Auspice ≥1.35.2

tsibley:
Auspice ≥1.35.2 includes a CLI which should be used to run Auspice going
forward, the invocation of which is slightly different than the previous
wrapper script.

This relies on the corresponding change in the base Docker image.


tsibley:
@barneypotter24 Would you test if this fixes the problem you ran into? You can test by checking out this `new-auspice` branch and then running:

```
./bin/nextstrain view --image nextstrain/base:branch-new-auspice ../path/to/some/jsons/
```

Thanks!

Issue #36, opened 2018-11-29
Commands to manipulate (download, delete, list) deployed datasets and narratives


jameshadfield:
The `nextstrain deploy` command works great to upload data to the S3 buckets. It struck me that we may want to download data as well, especially for S3 buckets where authentication is required. (Maybe this would entail renaming `deploy`). We do this in auspice via https://github.com/nextstrain/auspice/blob/master/scripts/get-data.sh, which is a bit hacky... wildcards would be really good if possible.

tsibley:
Hmm. This seems like scope creep to me. The `nextstrain` command isn't meant to be a general-purpose S3 manipulation tool: `aws s3` exists for that (and supports wildcards).

Note that S3 is just one possible destination for `nextstrain deploy` (albeit the only destination possible right now); future supported destinations might use SFTP, SCP, git, WebDAV, or other upload mechanisms. The intent is for `nextstrain deploy` to do what needs to be done to put locally built files at a location reachable by a deployed Nextstrain instance.

jameshadfield:
with the implementation of private S3 buckets, we will need the ability to download / remove files (right now we can only do this via the AWS console).

tsibley:
Ok, it sounds like we should re-envision the `nextstrain deploy` command.

What about something like the following, where `deploy` becomes `dataset upload`:

```
nextstrain dataset upload s3://nextstrain-data auspice/*.json
nextstrain dataset download s3://nextstrain-data/zika_tree.json some/dir/
nextstrain dataset delete s3://nextstrain-data/zika*

# Deploy will be an alias for `dataset upload`, so this still works, but asks
# you to start using the new command.
nextstrain deploy s3://nextstrain-data *.json
```

The `dataset delete` command will also take care of invalidating the CloudFront cache if necessary, the same way `deploy` does now.

trvrb:
This seems very reasonable to me. I would find `nextstrain dataset download` useful in day-to-day work with flu. 

However, `dataset` seems slightly non-perfect semantically, but I don't have a better suggestion.

I'm 50/50 on whether I prefer:

```
nextstrain dataset upload s3://nextstrain-data auspice/*.json
nextstrain dataset download s3://nextstrain-data/zika_tree.json some/dir/
nextstrain dataset delete s3://nextstrain-data/zika*
```

or 

```
nextstrain retrieve s3://nextstrain-data auspice/*.json
nextstrain deploy s3://nextstrain-data/zika_tree.json some/dir/
[...]

PR #40, opened 2019-02-05
Keep zipped S3 file and CloudWatch logs

trvrb:
When running a series of `--aws-batch` jobs if one would fail, then I couldn't review logs, which was frustrating. Also, when running a series of `--aws-batch` jobs, I would often want the ability to redownload zipped file from S3. Rather than deleting on download, I think better to add an automatic removal of old files from the nextstrain-jobs bucket.

### DIFF
--- a/nextstrain/cli/runner/aws_batch/__init__.py
+++ b/nextstrain/cli/runner/aws_batch/__init__.py
@@ -169,21 +169,17 @@ def run(opts, argv, working_volume = None) -> int:
         s3.download_workdir(remote_workdir, local_workdir)
 
 
-    # Remove remote results.
-    print_stage("Cleaning up S3")
-
-    remote_workdir.delete()
-
-    print("deleted:", s3.object_url(remote_workdir))
+    # Don't remove remote results
+    # print_stage("Cleaning up S3")
+    # remote_workdir.delete()
+    # print("deleted:", s3.object_url(remote_workdir))
 
 
-    # Remove log stream.
-    if job.log_stream:
-        print_stage("Cleaning up logs")
-
-        logs.delete_stream(job.log_stream)
-
-        print("deleted log stream:", job.log_stream)
+    # Don't remove log stream
+    # if job.log_stream:
+    #     print_stage("Cleaning up logs")
+    #     logs.delete_stream(job.log_stream)
+    #     print("deleted log stream:", job.log_stream)
 
 
     # Exit with the job's exit code, or assume success


jameshadfield:
Potentially related to #36

tsibley:
@trvrb Maybe we can chat about this more in person tomorrow or Friday, but some thoughts:

I have no major qualms with leaving the logs and results zip files lying around for now. Both the log group and S3 bucket configurations already include policies to automatically delete orphaned logs/files after 30 days.

My plan from the start has been to support resuming of remote builds in a future version, which would include keeping logs and results around + some friendly way to re-download logs and results using the `nextstrain` command. In the meantime, since that feature doesn't exist yet, would you review logs by logging into the AWS Console? Or using the `aws` CLI? What about the zip file of results? I'm trying to understand what the interim usage might look like.

Could you explain more about the situation in which you couldn't review logs? The logs should be streamed to `nextstrain build`'s stdout, so should be in scrollback or where ever stdout is redirected to. Was this not working? If so, that's a CLI bug.

If we're making this change, then the code for the cleanups should just be deleted, not commented out.

trvrb:
I realize this is a work around. Here is the scenario:

1. I run `python batch.py` (https://github.com/nextstrain/seasonal-flu/blob/master/batch.py) to farm out 16 `--aws-batch` jobs that run in the background.

2. I monitor separate jobs via the AWS batch console. This lets me see that 16 jobs are running and gives me a link to the CloudWatch logs. This all works while the job is running, but after the job is finished I can't review the CloudWatch log. This is worse for failed jobs, where I can't review the log file to find out why it failed.

3. In farming out these 16 jobs, the eventual S3 download is not always working. I've fallen back to using the S3 console to download these zip files.

tsibley:
Ok, thanks for the details. I've made [this branch](https://github.com/nextstrain/cli/compare/keep-logs-and-results-zip) and will merge and make a new CLI release when it passes tests on Travis.

> 2. I monitor separate jobs via the AWS batch console. This lets me see that 16 jobs are running and gives me a link to the CloudWatch logs. This all works while the job is running, but after the job is finished I can't review the CloudWatch log. This is worse for failed jobs, where I can't review the log file to find out why it failed.

Yeah, when I wrote the CLI my intention was that folks wouldn't really be using the AWS console to monitor jobs or logs. I expected that people who wanted to save the logs would redirect the command's output to a file (possibly using `tee` so that they could see it on the screen at the same time).

Note that the `batch.py` script redirects each build's output to a log file in `logs/live_flu_{lineage}_{resolution}.txt`, so you should be able to always find output there locally. You can follow all builds at once with `tail -f logs/*`, or change the path to specify a single file or some subset to follow.

> 3. In farming out these 16 jobs, the eventual S3 download is not always working. I've fallen back to using the S3 console to download these zip files.

Hmm, that's strange and not right! It sounds like I should try running some batch seasonal builds too to see the problem for myself. Have you done any debugging of this yet, or do you have an example log file from such a job?

tsibley:
Released as nextstrain-cli 1.9.0.

trvrb:
Thank you @tsibley!

>Note that the `batch.py` script you linked to redirects each build's output to a log file in `logs/live_flu_{lineage}_{resolution}.txt`, so you should be able to always find output there locally. You can follow all builds at once with `tail -f logs/*`, or change the path to specify a single file or some subset to follow.
Separately, this logging also wasn't working for me. I'd get the initial `zipped...` calls, but the log file wasn't filling as it should. I think this `batch.py` script could be improved. We should discuss at some point.

>Hmm, that's strange and not right! It sounds like I should try running some batch seasonal builds too to see the problem for myself. Have you done any debugging of this yet, or do you have an example log file from such a job?
I believe this is happening when I batch jobs, close my computer and open again an hour or two later when jobs are finished, but where I don't immediately have internet.

PR #43, opened 2019-02-22
runner.aws_batch: Specify requested vCPU and memory

### LINKED ISSUE #42: AWS Batch runner should accept --cpus and --memory options
These options would modify the job submission, overriding any defaults in the job definition. Coupled with a cascade of compute environments defined for the job queue (or an auto-scaling optimal compute environment), this lets users who knows what their builds needs ask for a certain amount of resources without the CLI needing to be tightly coupled to AWS Batch config. This also makes it clear in the command-line invocation the interaction between resources requested and snakemake's `-j` option.

@trvrb would like this for seasonal flu builds.

trvrb:
This passes `--aws-batch-cpus` and `--aws-batch-memory` to `containerOverrides`. This also moves defaults to the CLI instead of AWS config.

This has been tested and appears to work just fine.

Resolves #42.

### DIFF
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -24,6 +24,13 @@ directory.
 [AWS Batch]: https://aws.amazon.com/batch/
 [`zika-tutorial/` directory]: https://github.com/nextstrain/zika-tutorial
 
+### Requesting resources
+
+By default AWS Batch jobs request 4 vCPUs and 7200MB of memory. Generally, when
+running Snamemake, the `--jobs` option should be matched to the requested number
+of vCPUs. These defaults can be overridden by specifying `--aws-batch-cpus` and
+`--aws-batch-memory`, for instance `--aws-batch-cpus=8` and
+`--aws-batch-memory=14800`.
 
 ## Configuration on your computer
 
@@ -75,7 +82,6 @@ or in the `~/.nextstrain/config` file
     s3-bucket = ...
 
 or passing the `--aws-batch-s3-bucket=...` option to `nextstrain build`.
-
 
 # Setting up AWS to run Nextstrain builds
 
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -27,10 +27,20 @@ directory.
 ### Requesting resources
 
 By default AWS Batch jobs request 4 vCPUs and 7200MB of memory. Generally, when
-running Snamemake, the `--jobs` option should be matched to the requested number
+running Snakemake, the `--jobs` option should be matched to the requested number
 of vCPUs. These defaults can be overridden by specifying `--aws-batch-cpus` and
 `--aws-batch-memory`, for instance `--aws-batch-cpus=8` and
-`--aws-batch-memory=14800`.
+`--aws-batch-memory=14800`. Alternatively, request can be specified in the
+`~/.nextstrain/config` file via
+
+    [aws-batch]
+    cpus = ...
+    memory = ...
+
+Or via setting environment variables `NEXTSTRAIN_AWS_BATCH_CPUS` and
+`NEXTSTRAIN_AWS_BATCH_MEMORY`. Note that requesting more CPUs or memory than
+available in a compute environment will result in a job that is queued but
+is never started.
 
 ## Configuration on your computer
 
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -24,23 +24,32 @@ directory.
 [AWS Batch]: https://aws.amazon.com/batch/
 [`zika-tutorial/` directory]: https://github.com/nextstrain/zika-tutorial
 
-### Requesting resources
+### Using and requesting resources
 
-By default AWS Batch jobs request 4 vCPUs and 7200MB of memory. Generally, when
-running Snakemake, the `--jobs` option should be matched to the requested number
-of vCPUs. These defaults can be overridden by specifying `--aws-batch-cpus` and
-`--aws-batch-memory`, for instance `--aws-batch-cpus=8` and
-`--aws-batch-memory=14800`. Alternatively, request can be specified in the
-`~/.nextstrain/config` file via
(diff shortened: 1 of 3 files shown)


doc/aws-batch.md
```diff
@@ -24,6 +24,13 @@ directory.
 [AWS Batch]: https://aws.amazon.com/batch/
 [`zika-tutorial/` directory]: https://github.com/nextstrain/zika-tutorial
 
+### Requesting resources
+
+By default AWS Batch jobs request 4 vCPUs and 7200MB of memory. Generally, when
+running Snamemake, the `--jobs` option should be matched to the requested number
+of vCPUs. These defaults can be overridden by specifying `--aws-batch-cpus` and
+`--aws-batch-memory`, for instance `--aws-batch-cpus=8` and
+`--aws-batch-memory=14800`.
```
tsibley: It would be good to mention that:

* AWS Batch may not be able to run your job if your configured compute environment doesn't allow for instances with the CPUs/memory you request.

* Defaults for the number of CPUs and amount of memory can be set using environment variables and/or the config file.

nextstrain/cli/runner/aws_batch/__init__.py
```diff
@@ -26,6 +26,13 @@
                  or config.get("aws-batch", "s3-bucket") \
                  or "nextstrain-jobs"
 
+DEFAULT_CPUS = os.environ.get("NEXTSTRAIN_AWS_CPUS") \
```
tsibley: The environment variable should be `NEXTSTRAIN_AWS_BATCH_CPUS` to match the other environment variable names and option names.

nextstrain/cli/runner/aws_batch/__init__.py
```diff
@@ -26,6 +26,13 @@
...
+            or config.get("aws-batch", "cpus") \
+            or "4"
+
+DEFAULT_MEMORY = os.environ.get("NEXTSTRAIN_AWS_MEMORY") \
```
tsibley: The environment variable should be `NEXTSTRAIN_AWS_BATCH_MEMORY` to match the other environment variable names and option names.

nextstrain/cli/runner/aws_batch/__init__.py
```diff
@@ -54,6 +61,20 @@ def register_arguments(parser) -> None:
         metavar = "<name>",
         default = DEFAULT_S3_BUCKET)
 
+    development.add_argument(
+        "--aws-batch-cpus",
+        dest    = "cpus",
+        help    = "Number of vCPUs to request for job",
+        metavar = "<name>",
```
tsibley: `metavar` should be `<count>` or `<number>` instead of `<name>`.

nextstrain/cli/runner/aws_batch/__init__.py
```diff
@@ -54,6 +61,20 @@ def register_arguments(parser) -> None:
...
+        default = DEFAULT_CPUS)
+
+    development.add_argument(
+        "--aws-batch-memory",
+        dest    = "memory",
+        help    = "Amount of memory in MB to request for job",
+        metavar = "<name>",
```
tsibley: `metavar` should be `<megabytes>` or `<number>` instead of `<name>`.

nextstrain/cli/runner/aws_batch/__init__.py
```diff
@@ -83,6 +104,8 @@ def run(opts, argv, working_volume = None) -> int:
             name       = run_id,
             queue      = opts.job_queue,
             definition = opts.job_definition,
+            cpus = int(opts.cpus),
+            memory = int(opts.memory),
```
tsibley: Adding `type = int` to the `add_argument` calls will ensure the types of `opts.cpus` and `opts.memory` are always correct and remove the need for the `int(…)` calls here. It will also cause argparse to throw an error early if someone provides a non-integer value.

tsibley: Maybe a minor thing, but when code is aligned for readability (on the `=` in this case), I like to keep any newly added code lined up too.

doc/aws-batch.md
tsibley: There's also a k → m typo in Snakemake.

nextstrain/cli/runner/aws_batch/__init__.py
trvrb: I see. I hadn't ever used `metavar` before. Just read the docs.

nextstrain/cli/runner/aws_batch/__init__.py
trvrb: This is much better.

trvrb: I hadn't noticed. Agreed it looks nicer.

trvrb:
Thanks for the pointers @tsibley. Very helpful. I agree about leaving defaults to be specified in the AWS Batch job description. Please take a look now.

tsibley:
Changes look good, thanks! The potential issue I brought up was indeed a problem when I tested it:

```
Parameter validation failed:
Invalid type for parameter containerOverrides.vcpus, value: None, type: <class 'NoneType'>, valid types: <class 'int'>
Invalid type for parameter containerOverrides.memory, value: None, type: <class 'NoneType'>, valid types: <class 'int'>
Job submission failed!
```

so I pushed a commit to fix it. I also pushed two other small commits which you may want to take a look at. If it all looks good, please feel free to merge.

trvrb:
Thanks for catching this Tom. The use of `**({ "vcpus": cpus } if cpus else {})` is really clever.

tsibley:
Nod. I prefer to use that sort of inline conditional when composing a larger data structure because it makes it easier to get a sense of the whole thing at once and in close context. Conditionals that mutate the data structure afterwards, e.g.

```
if cpus:
    job_submission["vcpus"] = cpus
```

don't provide that same closely-bound context, but sometimes they're unavoidable.

PR #45, opened 2019-02-22
Include pylintrc rules

trvrb:
This was copied over directly from augur/. It removes PyLint checking for whitespace, among some other semantic checks.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,550 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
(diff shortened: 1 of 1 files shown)


tsibley:
What's the goal here?

trvrb:
@tsibley ---

I hugely benefit from having a linter running when coding. And it would be useful to keep augur Python code linted similarly to cli Python code. Turning this on didn't flag much. A couple unused imports. Some whitespace (of course). Preference by pylint for default `return` over `return` embedded in `else`.

trvrb:
It sounds like linting isn't preferred by cli devs. Please go ahead and close if you'd like @tsibley.

Issue #50, opened 2019-08-28
recommend adding `--native` flag to `nextstrain view`


alliblk:
Since there's issues running the CLI in a docker container on Windows subsystem for Linux (WSL), there's this nice functionality where you can specify within `nextstrain build` with flag `--native` that you want to run the build with the locally installed `augur` and `auspice` rather than using those utilities within a container.

I'm opening this issue because I think that that this `--native` flag should be added to `nextstrain view` as well. While I understand that the nextstrain command is just a wrapper, and I'm able to view my build with `auspice view`, I think that switching from the nextstrain-cli command notation to augur/auspice command notation will be confusing for WSL users less familiar with running nextstrain and the differences between and the local and the container ways of using it.

Also I know this is a big ask (I'm sorry Tom), but is there any way that this can be done before I head to the DRC?

PR #51, opened 2019-08-28
allow view to be run with --native

jameshadfield:
cc @alliblk does this make things work re: #50

### DIFF
--- a/nextstrain/cli/command/view.py
+++ b/nextstrain/cli/command/view.py
@@ -14,7 +14,7 @@ import re
 import netifaces as net
 from .. import runner
 from ..argparse import add_extended_help_flags
-from ..runner import docker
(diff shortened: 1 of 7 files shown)


tsibley:
@jameshadfield I've pushed up the completed support for `nextstrain view --native`. For full support, it pairs with a small change to Auspice that I submitted as nextstrain/auspice#781.

@alliblk Would you test this again locally from the branch?

tsibley:
> One thing that I did notice is that `build --help` still gives `--native` as an option. I'm assuming that's because it truly still is an option, but bringing it up incase `--native` was deprecated.

Ah, I didn't add `--native` vs. `--docker` auto-detection yet. I just fixed up `view --native` to work in more cases.

PR #52, opened 2019-08-30
Default environment (runner) with `check-setup --set-default`

tsibley:
@alliblk When you get a a chance tomorrow, would you test out this `default-runner` branch? You should be able to run `./bin/nextstrain check-setup --set-default` and see output at the bottom like:

    Setting default environment to docker

If docker support isn't found but native is, it should say "native" instead.

It'd be great to test it on the Windows machine, but one way to also test this out on your Mac is by stopping Docker using its menu in the Mac menu bar. Once a default is set, you should be able to run commands like `nextstrain build` and `nextstrain view` without specifying any runner/environment. You should also see the command's `--help` output mention which of `--docker` or `--native` is the default.

### DIFF
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -19,6 +19,7 @@ Three environments are supported, each of which will be tested:
 
 from functools import partial
 from textwrap import indent
+from .. import config
 from ..types import Options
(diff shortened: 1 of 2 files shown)


tsibley:
@alliblk's going to test this after release, so merging now.

PR #54, opened 2019-09-17
Detached builds on AWS Batch

### LINKED ISSUE #47: facilitate detached builds on AWS
I'd like to be able to run many builds simultaneously on the cloud, and then interrogate the results. The solution described in this issue doesn't fully achieve this, but makes the process smoother.

Currently the `--aws-batch` functionality is designed to stream back logs to the terminal. Quitting the program while running will stop the job running (on EC2), and while you could background the process and have it keep running, this is not ideal for this use case.

Running `netstrain build --aws-batch --detached ${build}` could trigger a batch job and simply return the job id. Subsequent functionality can be built in to query the progress of this job id / download results etcetera.

Note that "detached" may not be the correct word. Feel free to suggest alternatives.

tsibley:
This adds a feature we've wanted from the beginning of the CLI.  By starting the build with `--detach`, the job is only submitted and then the CLI prints the command necessary to `--attach` to the job later.  This command includes the job id and can be used as many times as desired.  It will even work on other computers or for other people, although you will need to modify the local build path to a directory of your choosing.

Jobs may also be interactively detached by pressing Control-Z.  Normally this would suspend a Unix process (which could be restarted with `fg` or `bg` or SIGCONT).  I think it's in the same spirit of suspending a local process to detach from the remote job instead.  It also parallels nicely with our existing Control-C job cancellation support.

There are currently no facilities to track job state locally or list outstanding jobs, but these features could be added later if it seems they'd be useful.  As it stands with this commit, a nice pattern for launching multiple detached jobs and picking them up later is:

    # Loop over several `nextstrain build` commands, appending the last
    # line to a shell script.
    nextstrain build --aws-batch --detach build-a/ | tail -n1 | tee -a pickup-jobs.sh
    nextstrain build --aws-batch --detach build-b/ | tail -n1 | tee -a pickup-jobs.sh
    …

    # Then, sometime later:
    bash pickup-jobs.sh

Some existing code was further indented, so this diff is bested viewed with `--ignore-all-space`.

Resolves #47.

I'd appreciate comments on the code but especially the interface/UX and functionality. I tagged several people because:

@jameshadfield, you've wanted this feature and filed #47.

@joverlee521 and @kairstenfay, I'd love to get your input on this and have more of the dev team familiar with the `nextstrain` CLI internals.

@trvrb and @joverlee521, I think we might want to use detached builds for seasonal-flu (and maybe seattleflu/augur-build?) going forward. This might simplify the current `batch.py` scripts and would also allow some sharing of job artifacts.


trvrb:
This is super exciting. I'll test with seasonal-flu as I'm running a bunch of these builds this week.

tsibley:
To make this a bit easier to test, I released a beta version to PyPi (1.14.0b1). This version won't be installed unless explicitly asked for. To install it, do something like:

```
python3 -m pip install --upgrade --pre nextstrain-cli
```

You may also need the `--user` flag if you installed it to your local user path instead of the system/a conda environment.

jameshadfield:
Thanks @trs -- I won't have time to dive into this for a week or two but this looks great

trvrb:
@tsibley --- Code looks good as far I can understand it. In testing things worked flawlessly for me. I was able to start a job, detach and attach to this job multiple times and in parallel. No bugs found.

This should be a major improvement to seasonal flu builds and to working with `--aws-batch` in general.

tsibley:
@trvrb Excellent, I'm glad to hear it! One thing I'd like to think about more (and why I didn't implement anything yet) is if there's a better way to handle starting batches of builds and picking them up again later, i.e. what `batch.py` in https://github.com/nextstrain/seasonal-flu/pull/39 does. If you have any thoughts on an interface for that, now or in the future, I'm all ears.

tsibley:
@jameshadfield If it's going to be a bit before you have a chance to review, any objection to using a review-after-merge instead of review-before-merge process here? I'd still welcome review comments and make any needed changes afterwards on a new PR.

(Ditto on review from @kairstenfay and @joverlee521.)

jameshadfield:
That's fine. Thanks for removing the thumbs down on https://github.com/nextstrain/cli/pull/54#issuecomment-532581864.

tsibley:
Great, thanks!

Since you mention it, know that I never saw the thumbs down which was ostensibly from me. Trevor mentioned it in a message on Friday while I was out of lab and when I opened the page this morning to look it wasn't there for me or him. I'm pretty confused about the situation, but hope you know that I wouldn't thumbs down a comment like that. :-/

trvrb:
>One thing I'd like to think about more (and why I didn't implement anything yet) is if there's a better way to handle starting batches of builds and picking them up again later, i.e. what batch.py in nextstrain/seasonal-flu#39 does. If you have any thoughts on an interface for that, now or in the future, I'm all ears.

`batch.py` no longer submits multiple jobs. I found it easier to define targets  and allocate resources based on target count. So, it's either submitting one big or one small job, rather than multiple small jobs. 

So, I don't have immediate need for starting batches and picking them up again later at this point.

tsibley:
> So, it's either submitting one big or one small job, rather than multiple small jobs.
> 
> So, I don't have immediate need for starting batches and picking them up again later at this point.

Ah, got it, I missed that.

PR #58, opened 2019-12-13
File management commands (upload, download, list, delete)

### LINKED ISSUE #36: Commands to manipulate (download, delete, list) deployed datasets and narratives

tsibley:
Adds the following commands:

    nextstrain remote upload
    nextstrain remote download
    nextstrain remote delete (rm)
    nextstrain remote list (ls)

Resolves #36. Commentary in the commit messages.

---

I know we talk about names a _lot_, but, well, names are important and tend to stick around once you use them.  Currently this PR uses _remote_. _Remote_ is our "can't think of anything else" default from discussion in #36.  It's fine, but coming from git it probably isn't intrinsically descriptive to our users.  It's a container that describes the location the contained commands act on in relative terms (in opposition to _local_).

_Files_ is another name I've been thinking of that's probably way more immediately recognizable to most people:  

    nextstrain files upload
    nextstrain files download
    nextstrain files delete (rm)
    nextstrain files list (ls)

It's a container that describes what is acted on by the contained commands.  When we talk about "JSONs" informally, we're eliding the more specific phrase "Auspice JSON files" or "Augur JSON files", both of which can reasonably and collectively called "Nextstrain JSON files", or just "Nextstrain files".  A slight downside I see here is the focus on individual files, which may get in the way semantically of moving to actions focused on [_datasets_](https://github.com/nextstrain/nextstrain.org/blob/seattleflu-group/docs/glossary.md#dataset), where each dataset is actually multiple files.  But we can have both file-based and dataset-based commands; they're not exclusive.

### DIFF
--- a/nextstrain/cli/command/deploy.py
+++ b/nextstrain/cli/command/deploy.py
@@ -4,9 +4,20 @@
 # Registering our own parser lets us preserve the original short description
 # and avoids introducing "upload" as a top-level command.
 
+from textwrap import dedent
 from .remote.upload import register_arguments, run, __doc__
 
 def register_parser(subparser):
     parser = subparser.add_parser("deploy", help = "Deploy pathogen build")
     register_arguments(parser)
     return parser
+
+def insert_paragraph(text, index, insert):
+    paras = text.split("\n\n")
+    paras.insert(index, dedent(insert))
+    return "\n\n".join(paras)
+
+__doc__ = insert_paragraph(
+    __doc__, 1, """
+    The `nextstrain deploy` command is an alias for `nextstrain remote upload`.
+    """)
--- a/nextstrain/cli/command/deploy.py
+++ b/nextstrain/cli/command/deploy.py
@@ -1,87 +1,12 @@
-"""
-Deploys a set of built pathogen JSON data files or Markdown narratives to a
-remote location.
+# This command, `nextstrain deploy`, is now an alias for `nextstrain remote
+# upload`.
+#
+# Registering our own parser lets us preserve the original short description
+# and avoids introducing "upload" as a top-level command.
 
-nextstrain.org, or standalone instances of Auspice, fetch the deployed JSON
-data files or Markdown narratives for display.
- 
- 
-Destinations
-------------
-
-Currently only Amazon S3 buckets (s3://…) are supported as the remote
-destination, but others can be added in the future.
-
-Destination URLs support optional path prefixes if you want your local
-filenames to be prefixed on the remote destination.  For example:
-
-    nextstrain deploy s3://my-bucket/some/prefix/ auspice/zika*.json
-
-will upload files named "some/prefix/zika*.json".
- 
- 
-Authentication
---------------
-
-Credentials for authentication should generally be provided by environment
-variables specific to each destination type.
-
-S3
---
-
-* AWS_ACCESS_KEY_ID
-* AWS_SECRET_ACCESS_KEY
-
-More information at:
-
-    https://boto3.readthedocs.io/en/latest/guide/configuration.html#environment-variables
-
-A persistent credentials file, ~/.aws/credentials, is also supported:
-
-    https://boto3.readthedocs.io/en/latest/guide/configuration.html#shared-credentials-file
- 
-"""
-
-from pathlib import Path
-from urllib.parse import urlparse
-from ..util import warn
-from ..remote import s3
-
-
-SUPPORTED_SCHEMES = {
-    "s3": s3,
-}
-
+from .remote.upload import register_arguments, run, __doc__
 
 def register_parser(subparser):
     parser = subparser.add_parser("deploy", help = "Deploy pathogen build")
-
-    # Destination
-    parser.add_argument(
-        "destination",
-        help    = "Deploy destination as a URL, with optional key/path prefix",
-        metavar = "<s3://bucket-name>")
-
-    # Files to deploy
-    parser.add_argument(
-        "files",
-        help    = "Files to deploy (e.g. JSON data files)",
-        metavar = "<file.json>",
-        nargs   = "+")
-
+    register_arguments(parser)
     return parser
-
-
-def run(opts):
-    url = urlparse(opts.destination)
-
-    if url.scheme not in SUPPORTED_SCHEMES:
-        warn("Error: Unsupported destination scheme %s://" % url.scheme)
-        warn("")
-        warn("Supported schemes are: %s" % ", ".join(SUPPORTED_SCHEMES))
-        return 1
-
-    remote = SUPPORTED_SCHEMES[url.scheme]
-    files  = [Path(f) for f in opts.files]
-
-    return remote.deploy(url, files)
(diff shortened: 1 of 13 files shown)


nextstrain/cli/remote/s3.py
```diff
@@ -78,33 +43,128 @@ def upload(local_files: List[Path], bucket, prefix: str) -> List[str]:
...
-    # Pass the original contents through gzip into memory
-    with GzipFile(fileobj = gzstream, mode = "wb") as gzfile:
-        shutil.copyfileobj(stream, gzfile)
+    except (NoCredentialsError, PartialCredentialsError) as error:
+        raise UserError("Error authenticating with S3: %s" % error) from error
+
+    # Find the bucket and ensure we have access and that it already exists so
+    # we don't automagically create new buckets.
+    try:
+        boto3.client("s3").head_bucket(Bucket = bucket.name)
```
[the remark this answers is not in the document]
tsibley: The S3 client doesn't require a region (because S3's API itself doesn't), so it doesn't matter. :-)

tsibley:
@joverlee521 Thanks for looking this over! I'm glad you learned a lot. :-)

nextstrain/cli/remote/s3.py
```diff
@@ -78,33 +43,128 @@ def upload(local_files: List[Path], bucket, prefix: str) -> List[str]:
...
+            return local_path / Path(obj.key).name
+        else:
+            return local_path
+
+    files = list(zip(objects, [local_file_path(obj) for obj in objects]))
+
+    for remote_object, local_file in files:
+        yield Path(remote_object.key), local_file
+
+        encoding = remote_object.content_encoding
```
jameshadfield: Downloading a file which doesn't exist (which I'll do a lot as I'll be meaning to use `--recursive` but forget the flag) results in a trace from this line. This should be caught & help printed. 

```
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/bin/nextstrain", line 11, in <module>
    load_entry_point('nextstrain-cli', 'console_scripts', 'nextstrain')()
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/__main__.py", line 10, in main
    return cli.run( argv[1:] )
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/__init__.py", line 56, in run
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/command/remote/download.py", line 73, in run
    for remote_file, local_file in downloads:
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/remote/s3.py", line 84, in download
    encoding = remote_object.content_encoding
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/boto3/resources/factory.py", line 339, in property_loader
    self.load()
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/boto3/resources/factory.py", line 505, in do_action
    response = action(self, *args, **kwargs)
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/botocore/client.py", line 320, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/botocore/client.py", line 623, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```

nextstrain/cli/remote/s3.py
```diff
@@ -78,33 +43,128 @@ def upload(local_files: List[Path], bucket, prefix: str) -> List[str]:
...
+    Delete the files deployed at the given remote *url*, optionally *recursively*.
+    """
+    bucket, path = split_url(url)
+
+    # Delete either all objects sharing a prefix or the sole object (if any)
+    # with the given key.  Both methods return the same results data structure.
+    objects = bucket.objects.filter(Prefix = path)
+
+    if recursively:
+        responses = objects.delete()
```
jameshadfield: I haven't tested this, but it looks as if this will allow (recursive) deletion of an entire bucket. I can see this being useful in certain cases (groups) but also very dangerous. Please consider requiring `path` to be set for recursive deletion.

nextstrain/cli/command/deploy.py
```diff
@@ -1,87 +1,12 @@
...
-from ..util import warn
-from ..deploy import s3
-
-
-SUPPORTED_SCHEMES = {
-    "s3": s3,
-}
+# This command, `nextstrain deploy`, is now an alias for `nextstrain remote
+# upload`.
+#
```
jameshadfield: Makes sense to not duplicate this code, but it results in a help message for `nextstrain deploy` which talks about `nextstrain remote upload`. The help should be amended so that it either states that this is an alias for `nextstrain remote upload` or to indicate that it's deprecated and users should move to `nextstrain remote upload`.

tsibley:
Thanks for the review, I really appreciate having someone else's hands and eyes on the code. Your suggestions are all good, and I'll implement them when I can next find the time. _Maaaybe_ slipped in this week.

> Question (not a request for changes): why are the paths expanded using different syntax to bash? Is this a URL thing, or a S3 thing, or something else? E.g. `nextstrain remote list s3://nextstrain-data/flu*` doesn't work as one might think (at least, comparing it with shell expansion).

No wildcard expansion is performed, just prefix matching. This is for two reasons:

1. S3 is not actually a filesystem, but a key-value store. Slashes aren't special in S3 keys, and their use is by convention only. (People are more used to filesystems.)

2. The Nextstrain dataset JSONs use underscores instead of slashes to represent nesting, and so prefix-based operations make sense for them too. For example, imagine two people (Bert and Ernie) in a shared group who are both running copies of the same build which produces `varicella.json`. They can cooperatively experiment with their own build parameters in the same group without stepping on each other's datasets by agreeing to upload using their name as a prefix. So Bert can upload his `varicella.json` using `nextstrain remote upload s3://nextstrain-sesame-research/bert_ varicella.json` and that would produce `bert_varicella.json` files viewable as `nextstrain.org/group/sesame-research/bert/varicella`.

When you provide a wildcard pattern in `nextstrain remote list s3://nextstrain-data/flu*`, the shell doesn't expand the `*` to anything because the path leading up to it doesn't exist. So `nextstrain` sees the literal `*` but doesn't do anything special with it. As there aren't any keys in the nextstrain-data bucket starting with the literal `flu*`, nothing is printed. (Some shells, or bash in a non-default config, will remove the `*` when it doesn't expand to anything, but this is atypical. You can test if this is the case by running `echo s3://nextstrain-data/flu*`.)

tsibley:
Ah, and I meant to add: There is definitely an argument to support wildcard matching against the object keys, and that could be added relatively easily in the future. It might be less surprising, but I wanted to consider it later rather than right now.

tsibley:
@jameshadfield I just pushed additional changes addressing your remaining review notes. Would you give them a look?

PR #59, opened 2019-12-13
hostenv: Include ID3C environment variables

trvrb:
These environment variables are needed to access ID3C during Seattle Flu Study Augur build.

### DIFF
--- a/nextstrain/cli/hostenv.py
+++ b/nextstrain/cli/hostenv.py
@@ -18,6 +18,11 @@ forwarded_names = [
     # Augur <https://nextstrain-augur.readthedocs.io/en/stable/envvars.html>
     "AUGUR_RECURSION_LIMIT",
     "AUGUR_MINIFY_JSON",
+
+    # ID3C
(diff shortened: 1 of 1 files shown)


tsibley:
Values for these are already added to our default AWS Batch job definition, so that should work. I'm guessing this is for when running the build locally?

Although I'd like to avoid adding more build-specific environment variables here, the alternate solutions I have in mind aren't ready/implemented/tested enough. So this seems fine.

trvrb:
I see now. That's what happened. I didn't immediately know why it was working with `--aws-batch` and not working locally. Local runs are highly desirable, but not strictly necessary. If you're okay merging it would be helpful for my work on this.

tsibley:
> If you're okay merging it would be helpful for my work on this.

Definitely. Merged. Would a new CLI release also be helpful here then?

trvrb:
I'm fine with `pip install -e .` for CLI. No hurry to release new version.

PR #67, opened 2020-03-28
Describe how set-default chooses and env

### LINKED ISSUE #53: [check-setup] Clarify help for --set-default
With release 1.11.1 we can now check the computer's hardware and assess what type of nextstrain running architectures are available (docker, native, or AWS). The thing that is nice about this is that you can have the same running commands (e.g. `nextstrain build <dir>`) on both a windows machine which can't use docker and a mac or some other machine that does allow docker.

You can also set a default run type to use, using the flag `--set-default` when running `nextstrain check-setup`. Currently when `nextstrain check-setup --help` is run, the help statement describes what `set-default` does, but does not specify the rank order of running methods that it will choose from when picking a default setting. This should be amended to be more clear that first choice for default is docker, followed by native.

Putting this issue up now to remember it, but I'm unlikely to get to this until after I'm back from the DRC.

mlarrousse:
Change the documentation on check-setup --set-default to clarify what will happen when the flag is used.

### Related issue(s)  
Fixes #53   

### Testing
Ran setup-checks and setup-checks -h.

### DIFF
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -30,7 +30,7 @@ def register_parser(subparser):
 
     parser.add_argument(
         "--set-default",
-        help   = "Set the default environment to use going forward based on the results of checking your setup",
+        help   = "Set the default environment to the first which passes check-setup. Checks run in the order: docker, native, aws-batch.",
         action = "store_true")
 
     return parser


tsibley:
Thanks @mlarrousse! I've merged this and it'll go out in the next release of the CLI sometime soon.

I did add one additional change to dynamically generate the runner order in the help text: https://github.com/nextstrain/cli/commit/44c73031dd0f59cc7825878a821b5c4cab1f3517

PR #71, opened 2020-04-03
Remove --tty flag from Docker invocation

### LINKED ISSUE #49: confusing behavior when a terminal window is closed with `nextstrain view` running
If you close a terminal window that was running `nextstrain view` then -- at least on MacOS -- the docker container continues to run and the port (normally 4000) still connects to the docker container. Combined with #48 this makes for some fun debugging. Is it possible to close the docker container in this case (as happens with ctrl+c), or is this not possible?

technicalciso:
fixes the issue on linux, should work on mac but needs be tested (I don't have one).  note that this will remove the colors in the term window - is it worth it? :)

fixes #49

### DIFF
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -79,7 +79,6 @@ def run(opts, argv, working_volume = None, extra_env = {}) -> int:
     return exec_or_return([
         "docker", "run",
         "--rm",             # Remove the ephemeral container after exiting
-        "--tty",            # Colors, etc.
         "--interactive",    # Pass through control signals (^C, etc.)
 
         # On Unix (POSIX) systems, run the process in the container with the same


tsibley:
Hi @technicalciso, thanks for the PR! Can you explain more about how this fixes #49? @jstoja [couldn't replicate](https://github.com/nextstrain/cli/issues/49#issuecomment-609085748) the issue a few days ago, so I'm interested if you were a) able to replicate and b) pinned down the reason why this fixes the issue. If you are able to replicate, can you paste your `docker version` output? Thanks!

tsibley:
Also, I'd much prefer to keep the colors and other tty features if we can. :-)

technicalciso:
hi @tsibley i am still able to replicate.  --tty connects the terminal's input to the docker's input (STDIN).  my guess is that when you close the terminal window, STDIN stays open in the docker and prevents it from closing. more specifically, i think _node_ is holding it open.  this behavior is expected once _readline_ is invoked (as per [docs](https://nodejs.org/api/readline.html): Node.js application will not terminate until the readline.Interface is closed because the interface waits for data to be received on the input stream). 

i am running Docker version 1.13.1, build cccb291/1.13.1 on CentOS 7.7

as for the colors, i may have spoken too soon.  the comments in Docker.py indicated "Colors, etc." next to the tty flag, but i checked and even without --tty, the colors don't actually seem to be impacted.

technicalciso:
hi @tsibley  checking to see if you were able to replicate?

tsibley:
Hi @technicalciso, thanks for checking in and diagnosing this issue. I've been slammed but looking into this is still on my todo list. Hopefully next week… (really need to take the weekend off/light!)

tsibley:
@technicalciso Ah, that information on node's readline is super helpful! It led me to [this observation](https://github.com/nextstrain/cli/issues/49#issuecomment-623032351).

That said, I'm still not sure that removing `--tty` is desirable. While some programs may always emit ANSI control sequences, others will modify their behaviour based on if stdout is a tty or not. I think removing `--tty` would affect this given:

```
$ docker run --rm --interactive --tty nextstrain/base test -t 1
# exits 0, e.g. stdout is a tty

$ docker run --rm --interactive nextstrain/base test -t 1
# exits 1, e.g. stdout is not a tty
```

Instead, I think a better solution might be to pass `--init` to Docker, so that PID 1 in the container does proper signal handling (which node is not doing). I'm not sure if this would have other unintended affects though, so would want to thoroughly test expected `nextstrain` invocations continue to work.

technicalciso:
@tsibley some processes will run this test, but most of the time it's because the process will behave differently if it's sending the output to a pipe rather than STDOUT. not sure if this will actually alter anything here? (the original bug only mentioned _nextrain view_, so that's all i tested)

also, looks like nexstrain view already runs with --init.  check the _opts.docker_args_ section in view.py which i think is where its launched from.  you can also confirm by running ps -ef |grep docker while its running which returns something like this:
/usr/bin/docker-current run --rm --tty --interactive --user=1000:1000 --volume=/home/me/nextstrain/zika-tutorial:/nextstrain/auspice/data --workdir=/nextstrain/auspice/data --env=AUGUR_RECURSION_LIMIT --env=AUGUR_MINIFY_JSON --env=AWS_ACCESS_KEY_ID --env=AWS_SECRET_ACCESS_KEY --env=AWS_SESSION_TOKEN --env=ID3C_URL --env=ID3C_USERNAME --env=ID3C_PASSWORD --env=RETHINK_HOST --env=RETHINK_AUTH_KEY --env=HOST --env=PORT **--init** --env=HOST=0.0.0.0 --publish=127.0.0.1:4000:4000 nextstrain/base auspice view --verbose --datasetDir=.

tsibley:
@technicalciso Ah, yes, you're totally right about `--init`. I was only looking in `runner/docker.py` not `command/view.py` which extends the default Docker options. Thanks for the correction!

> some processes will run this test, most of the time it's because the process will behave differently if it's sending the output to a pipe rather than STDOUT

Right. Some programs won't emit colored output if they think stdout is a pipe or file and not a tty. I believe that's why the `--tty` line is commented with `Colors, etc`. For example, consider the difference between these two commands:

```
docker run --rm --interactive       nextstrain/base ls /
docker run --rm --interactive --tty nextstrain/base ls /
```

While we don't really care about `ls` in this context, many other programs we do care about behave similarly to `ls`, such as `snakemake`. So I'm inclined not to remove `--tty` wholesale. It might make sense to remove it just for `nextstrain view`, though.

technicalciso:
@tsibley take a look at the comment i posted [here](https://github.com/nextstrain/cli/issues/49#issuecomment-624382829). if you think people are using the RHEL docker and still want to fix this for nextstrain view only i can work on it.

PR #79, opened 2020-05-20
Make the logging thread more resilient.

ttung:
On occasion, there would be ClientErrors caused by InvalidSignatureException.  This is likely caused by laggy networking (I was uploading a dataset to S3 simultaneously).  This crashes the logging thread.

This PR counts the number of successful and failed requests, and if there have been successful requests, it ignores the failed requests.  If there have never been any successful requests, then it raises an exception after 10 failed requests.

### Testing
Ran the zika dataset.

### DIFF
--- a/nextstrain/cli/runner/aws_batch/logs.py
+++ b/nextstrain/cli/runner/aws_batch/logs.py
@@ -3,11 +3,12 @@ Log handling for AWS Batch jobs.
 """
 
 import threading
-from typing import Callable, Generator, Mapping, MutableSet
+from typing import Callable, Generator, MutableSet
 from ... import aws
 
 
 LOG_GROUP = "/aws/batch/job"
+MAX_FAILURES = 10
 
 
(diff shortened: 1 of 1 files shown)


tsibley:
Thanks for this PR, @ttung! I really appreciate making the CLI more robust.

Whenever I see code that catches StopIteration, I wonder about alternate ways of structuring it to avoid what feels like an implementation detail to me. In this case, I tried out a variation of your PR in this repo's [`resilient-logging` branch](https://github.com/nextstrain/cli/compare/resilient-logging). You can see the [comparison between our branches](https://github.com/ttung/cli/compare/ttung/resilient-logging...nextstrain:resilient-logging). The gist is that I preserved the `for` loop, which I think keeps a lot of clarity here, and, to make up for expanded exception catchment of the new `try` block, narrowed the `except` clause back down to the expected botocore errors.

Thoughts? Does this work for you? I want to make sure it does. :-)

ttung:
Sure, this works for me. :)

tsibley:
Merged and released (along with #80) as 1.16.7.

Issue #75, opened 2020-05-01
[aws batch] Gracefully handle network-level errors during uploading


jameshadfield:
**Current Behavior**  

I'm getting this error frequently when running ncov builds on aws-batch -- only around 1/3 of my attempts to run the command succeed. It's related to uploading the large (400+Mb) repo which is needed to run SARS-CoV-2 builds currently. The internet is not down, so it's not the same as #55.

```
$ nextstrain build --aws-batch --detach ...
...
zipped: /Users/naboo/github/nextstrain/ncov/Snakefile_Priorities

Traceback (most recent call last):
  File "/Users/naboo/miniconda3/lib/python3.7/site-packages/urllib3/connection.py", line 157, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw
  File "/Users/naboo/miniconda3/lib/python3.7/site-packages/urllib3/util/connection.py", line 61, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/Users/naboo/miniconda3/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno 8] nodename nor servname provided, or not known

During handling of the above exception, another exception occurred:
...
  File "/Users/naboo/miniconda3/lib/python3.7/site-packages/botocore/httpsession.py", line 283, in send
    raise EndpointConnectionError(endpoint_url=request.url, error=e)
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: <redacted>
```

This is the nextstrain-cli function which triggers it:
```
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/runner/aws_batch/__init__.py", line 121, in run
    remote_workdir = s3.upload_workdir(local_workdir, bucket, run_id)
  File "/Users/naboo/github/nextstrain/cli/nextstrain/cli/runner/aws_batch/s3.py", line 70, in upload_workdir
    remote_workdir.upload_fileobj(tmpfile)
```

**Expected behavior**  
Ability to upload large build directories despite less-than-ideal network upload speeds.

**Additional context**  
nextstrain.cli 1.16.5

tsibley:
My intuition tells me this is most likely a network- or host-level problem that the CLI can't solve. We might be able to provide better messaging by catching the error, though I'm not sure yet what that message would be.

tsibley:
[...]

PR #80, opened 2020-05-20
Add retries for the S3 resource

ttung:
This adds 3 retries.  I'm not sure if this is excessive, but I do see sporadic failures in uploading the zip.  The better solution is to do a multipart upload with retries (since it wouldn't need to start all over), but this should be a UX improvement.

### Related issue(s)  
Fixes https://github.com/nextstrain/cli/issues/75

### Testing
Ran the zika tutorial

### DIFF
--- a/nextstrain/cli/runner/aws_batch/s3.py
+++ b/nextstrain/cli/runner/aws_batch/s3.py
@@ -4,6 +4,7 @@ S3 handling for AWS Batch jobs.
 
 import binascii
 import boto3
+from botocore.config import Config
 from botocore.exceptions import ClientError
 from calendar import timegm
 from os import utime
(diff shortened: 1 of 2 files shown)


nextstrain/cli/runner/aws_batch/s3.py
```diff
@@ -187,14 +188,15 @@ def bucket(name: str) -> S3Bucket:
     """
     Load an **existing** bucket from S3.
     """
-    bucket = boto3.resource("s3").Bucket(name)
+    config = Config(retries={'max_attempts': 3})
+    s3_resource = boto3.resource("s3", config=config)
 
     try:
-        boto3.client("s3").head_bucket(Bucket = bucket.name)
+        s3_resource.meta.client.head_bucket(Bucket=name)
```
ttung, line 195: This is an [official way](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/clients.html#creating-clients) of getting a client instance.

tsibley:
Thanks for this improvement @ttung!

Issue #77, opened 2020-05-02
[build] Add options for CPUs/cores/threads and memory limits shared by all runners


tsibley:
The goal would be to deprecate `--aws-batch-cpus` and `--aws-batch-memory` in favor of generic options supported by all runners (or ignored if not applicable). If the `--exec` program is `snakemake` (the default), then these values would also be automatically passed into it.

An equivalency table of how these proposed new options would map to existing options:

proposed | existing
--- | ---
`nextstrain build --cpus 2 --memory 1GiB --docker .` | `nextstrain build --docker . --cores 2 --resources mem_mb=1024`
`nextstrain build --cpus 2 --memory 1GiB --aws-batch .` | `nextstrain build --aws-batch --aws-batch-cpus 2 --aws-batch-memory 1024 . --cores 2 --resources mem_mb=1024`
`nextstrain build --cpus 2 --memory 1GB --aws-batch .` | `nextstrain build --aws-batch --aws-batch-cpus 2 --aws-batch-memory 954 . --cores 2 --resources mem_mb=954`

`--memory` would accept 1024-based units like `MiB` and `GiB` as well as 1000-based units `MB`, and `GB`. The default would be `MB` if no unit is specified.

I'm not sure if `--cpus` should be `--cpus`, `--cores`, `--threads` or something else. All terms are somewhat ambiguous due to overloaded meanings.

Related to #65, which is a more specific issue that would be solved by this more general one.

Providing these general options allows for a cleaner, simpler build invocation regardless of build environment and, in the future where workflow executors other than Snakemake are supported, cleaner integration with their ways of specifying resources too.

PR #81, opened 2020-05-21
[build] Add --cpus and --memory options

### LINKED ISSUE #65: [aws batch] Runner should automatically provide snakemake with the appropriate `--cores N` option
If the AWS Batch runner is executing `snakemake`, it should provide a default `--cores N` / `--jobs N` option where N defaults to the number of CPUs in the Batch job definition (which [may be overridden by the environment or CLI options](https://github.com/nextstrain/cli/blob/015ab9979ef4d38b7ce17723e706d01f0b29b690/doc/aws-batch.md#using-and-requesting-resources)).

This will let the Nextstrain build use all the available resources by default, instead of defaulting to a single core.

Any user-provided `--cores`, `--jobs`, or `-j` option shouldn't be overridden.

tsibley:
Limits containerized (Docker, AWS Batch) builds to the given amounts and
when applicable, informs Snakemake's resource scheduler and/or the AWS
Batch instance size selection.

Resolves #65 and #77.

---

See commit messages for more details.

### DIFF
--- a/nextstrain/cli/command/build.py
+++ b/nextstrain/cli/command/build.py
@@ -23,9 +23,11 @@ is as easy as possible.  It also lets us more seamlessly make environment
 changes in the future as desired or necessary.
 """
 
+import re
+from textwrap import dedent
 from .. import runner
 from ..argparse import add_extended_help_flags
-from ..util import warn
+from ..util import byte_quantity, warn
 from ..volume import store_volume
 
 
@@ -53,6 +55,25 @@ def register_parser(subparser):
                "Currently only supported when also using --aws-batch.",
         metavar = "<job-id>")
 
+    parser.add_argument(
+        "--cpus",
+        help    = "Number of CPUs/cores/threads/jobs to utilize at once.  "
+                  "Limits containerized (Docker, AWS Batch) builds to this amount.  "
+                  "Informs Snakemake's resource scheduler when applicable.  "
+                  "Informs the AWS Batch instance size selection.",
+        metavar = "<count>",
+        type    = int)
+
+    parser.add_argument(
+        "--memory",
+        help    = "Amount of memory to make available to the build.  "
+                  "Units of b, kb, mb, gb, kib, mib, gib are supported.  "
+                  "Limits containerized (Docker, AWS Batch) builds to this amount.  "
+                  "Informs Snakemake's resource scheduler when applicable.  "
+                  "Informs the AWS Batch instance size selection.  ",
+        metavar = "<quantity>",
+        type    = byte_quantity)
+
     # Positional parameters
     parser.add_argument(
         "directory",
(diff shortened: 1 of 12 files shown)


nextstrain/cli/command/build.py
```diff
@@ -77,4 +98,94 @@ def run(opts):
...
+                # So for now, if folks are specifying their own --resources,
+                # they'll also need to explicitly provide "mem_mb", which may mean
+                # repeating a previous --memory argument they provided us.
+                #   -trs, 20 May 2020
+                warn(dedent("""
+                    Warning: The explicit %s option passed to Snakemake prevents
+                    the Nextstrain CLI from automatically providing a "mem_mb" resource
+                    based on its --memory option.  This may or may not be what you expect.
+                    """ % (snakemake_opts["--resources"][0],)))
+
```
kairstenfay, line 135: Could we utilize the fact that `opts.exec == "snakemake"` is the one "happy" conditional here to reduce the indentation of this code block?

nextstrain/cli/util.py
```diff
@@ -213,3 +213,56 @@ def resolve_path(path: Path) -> Path:
...
+    >>> byte_quantity("2Kb")
+    2000
+    >>> byte_quantity("2 kib")
+    2048
+    >>> byte_quantity("1.5GB")
+    1500000000
+    >>> byte_quantity("1024")
+    1024
+    """
+    match = re.search(r'(?xi)^ (\S+?) \s* ((?:(?:[kmg]i?)?b)?) $', quantity.strip())
```
kairstenfay, line 248: What is going on here :laughing: 
I understand the second half, `((?:(?:...` : the code is doing a look ahead for k/m/g (i)(b) suffixes without matching on them. But I am struggling to understand the function of `(?xi)^ (\S+?)`.

nextstrain/cli/command/build.py
tsibley: Hmm, good thought. I think that'd require repeating the `runner.run` call though?

```python
if opts.exec != "snakemake":
   return runner.run(opts, working_volume = opts.build, cpus = opts.cpus, memory = opts.memory)

# …bunch of arg manipulation here, then…

return runner.run(opts, working_volume = opts.build, cpus = opts.cpus, memory = opts.memory)
```

Not sure I like repeating that, as it seems easy to miss the two places when updating args.

kairstenfay: Oh, right. That would cause unnecessary duplication. I suppose there's no  good way around it.

nextstrain/cli/util.py
tsibley: `(?:…)` is a non-capturing group, not a look-ahead. Those non-capturing groups are surrounded by a capturing group `(…)`.

`(?xi)` sets the flags for the regex. It must be the first thing ("regex atom") when its used. `x` is the `re.X` or `re.VERBOSE` flag. `i` is the `re.I` or `re.IGNORECASE` flag. I'm used the short flags from decades of using them, but perhaps I should expand them for clarity here and move them into arguments to `re.search` instead of part of the pattern. (I like them as part of the pattern because they tell me upfront how to read the pattern, instead of coming afterwards, but this is a minor thing overall.)

`(\S+?)` is not an optional match, but a shortest ("ungreedy") match of one or more non-whitespace chars. `?` modifies a quantifier, in this case `+`, to change its behaviour from longest ("greedy") match to shortest ("ungreedy") match. This ensures that the `\S+` atom doesn't eat up the unit specifier in a string without spaces, like `18gb`.

Would this be better?

```python
re.search(
    r"""
    ^                           # start of string
    (?P<value> \S+? )           # the numeric value: shortest match of non-whitespace chars
    \s*                         # optional whitespace between the value and the unit
    (?P<unit>
     (?:
      (?: [kmg]i? )? b)? )      # the unit: [kmg]b, [kmg]ib, b, or nothing
    $                           # end of the string
    """,
    quantity.strip(),
    re.VERBOSE | re.IGNORECASE)
```

or maybe

```python
re.search(
    r"""
    ^                           # start of string
    (?P<value> \S+? )           # the numeric value: shortest match of non-whitespace chars
    \s*                         # optional whitespace between the value and the unit
    (?P<unit>                   # the unit:
          [kmg]b                #    kb, mb, gb
        | [kmg]ib               #    kib, mib, gib
        | b                     #    b
    )?                          # or none of the above, in which case we assume bytes.
    $                           # end of the string
    """,
    quantity.strip(),
    re.VERBOSE | re.IGNORECASE)
```

joverlee521: Testing on https://regex101.com it currently matches strings like `18kmb` with `18k` being the `value`, but I guess `float(value)` would raise a ValueError in this case?

kairstenfay: I thought I understood regex fairly well until this PR. I did not even realize python had regex flags you could set at the beginning of the expression.

Thank you for the explanations. Jover also pointed me to regex101.com which I will use in the future for understanding complex expressions. 

As for documenting the pieces of the regex expression you wrote, I think the usefulness is pretty subjective. For example, `^` and `$` are very straightforward for me and don't need an explanation, but others may disagree. I personally think explaining the more complex pieces of the regex is useful. Or I could just turn to regex101.com in the future.

tsibley: Yeah, it relies on the float conversion to raise errors with the value part. This seemed easier than representing everything understood by `float()` as a regex, e.g. `12e3 mb` or `.5gb`. Maybe it should be stricter though.

tsibley: I repushed this with something I think is at least a little more clear.

nextstrain/cli/command/build.py
```diff
@@ -77,4 +98,94 @@ def run(opts):
...
+    >>> parse_snakemake_args(["-j8"])
+    {'--cores': ['-j'], '--resources': []}
+
+    >>> parse_snakemake_args([])
+    {'--cores': [], '--resources': []}
+    """
+    # XXX TODO: Consider using a small ArgumentParser() for this in the
+    # future, when we can require Python 3.7 and use parse_intermixed_args().
+    #   -trs, 20 May 2020
```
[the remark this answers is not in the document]
tsibley: Hmm, it's not clear to me any more why something like a partial ArgumentParser for Snakemake's options applied as `parse_known_args(extra_exec_args)` wouldn't be sufficient. I suspect I tried to this path and couldn't get it to work due to option/argument intermixing, but not sure.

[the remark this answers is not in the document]
tsibley: WFM!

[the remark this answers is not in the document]
victorlin: f84e0791b1bf031f768cbcb8075970ddf70bb656

PR #82, opened 2020-05-22
[AWS Batch] Support overriding the container image used

### LINKED ISSUE #57: AWS Batch runner does not support runtime-configurable Docker images
This limitation is because the `--aws-batch` runner (`nextstrain.cli.runner.aws_batch`) relies on a static job definition created manually during AWS Batch setup ([as documented](https://github.com/nextstrain/cli/blob/master/doc/aws-batch.md)). The job definition hardcodes an image, and individual job instances based on that definition are not allowed to override the image (unlike other job properties like CPU, memory, command line, etc).

The  `--aws-batch` runner should allow a Docker image to be selected by the user at runtime, not just setup, using the same mechanisms supported by the `--docker` runner. Specifically, `--aws-batch` should use the image passed in via `--image` or the default Docker image (configured via a config file or environment var).

To accomplish this, the `--aws-batch` runner will have to create a new job definition before job submission if no job definition already exists for the desired image. The new job definition should be based on the default job definition, using it as a template for all properties except the image. The help message for `--aws-batch-job` should be updated to document that it may be used this way.

As a rough map of where to start for the person taking on this issue, I [annotated the places](https://gist.github.com/tsibley/fd0b878302f2f176cb51b47e8fa58b91) which need new code to support this functionality.

tsibley:
Job definitions are dynamically created as needed based on the base
definition specified by `--aws-batch-job`.  This is necessary to allow
runtime-configurable images since, unlike CPUs, memory, env vars, etc, a
submitted job may not directly override the container image set in the
job definition.

Resolves #57.  Implementation originally started as PR #73.

Co-authored-by: Julien BORDELLIER

### DIFF
--- a/nextstrain/cli/runner/aws_batch/jobs.py
+++ b/nextstrain/cli/runner/aws_batch/jobs.py
@@ -2,9 +2,15 @@
 Job handling for AWS Batch.
 """
 
+import re
+from botocore.exceptions import ClientError
+from copy import deepcopy
+from operator import itemgetter
 from time import time
 from typing import Callable, Generator, Iterable, Mapping, List, Optional
 from ... import hostenv, aws
+from ...errors import UserError
+from ...util import split_image_name
 from . import logs, s3
 
 
@@ -139,6 +145,7 @@ class JobState:
 
 
 def submit(name: str,
+           image: str,
            queue: str,
            definition: str,
            cpus: Optional[int],
@@ -156,7 +163,7 @@ def submit(name: str,
     submission = batch.submit_job(
         jobName = name,
         jobQueue = queue,
-        jobDefinition = definition,
+        jobDefinition = override_definition(definition, image),
         containerOverrides = {
             "environment": [
                 {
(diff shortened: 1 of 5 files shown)


nextstrain/cli/runner/aws_batch/jobs.py
```diff
@@ -219,16 +226,116 @@ def forwarded_environment() -> List[dict]:
...
+            del derived_definition[key]
+
+        batch.register_job_definition(**derived_definition)
+    else:
+        derived_image = derived_definition["containerProperties"]["image"]
+
+        assert split_image_name(derived_image) == split_image_name(image), (
+            "Expected job definition «%s», derived from «%s», to use image «%s», "
+            "but it uses «%s» instead"
+            % (derived_definition_name, base_definition_name, image, derived_image))
```
joverlee521, line 298: Just out of curiosity, why use `%s` instead of `.format()`? (Looks like `%s` is used for string formatting for most of this repo)

tsibley: Astute observation! I find `.format()` slightly clunkier than printf-style most of the time, so tend not to use it. Just pure preference! I prefer f-strings the most, but those aren't supported until Python 3.6 and this project currently documents support for 3.5. Python 3.5 was still somewhat common on Linux computers when I was first starting this project in 2018. When this project moves to 3.6 or higher as a minimum Python version, I'd like to move to f-strings.

Issue #90, opened 2020-06-17
[remote] Support Nextstrain Groups as remote destinations using nextstrain.org credentials


tsibley:
Currently the `nextstrain remote` family of commands only support `s3://` remote destinations, but they're designed with the intention to support other kinds of remote destinations in the future. The future is now. :-)

One of the pain points with Nextstrain Groups is that using `nextstrain remote` to deploy, list, and manage datasets in your group requires setting up AWS credentials and knowing the name/path of the S3 bucket. Instead, we'd like to support using nextstrain.org credentials (see #89) and a custom Nextstrain Groups destination syntax, thus decoupling the deploy process from our AWS implementation.

What should the destination syntax be? I can imagine the following, in my rough order of preference:

```
nextstrain remote upload nextstrain.org/groups/blab/ auspice/sars-like-cov*.json
nextstrain remote upload nextstrain:groups/blab/ auspice/sars-like-cov*.json
nextstrain remote upload groups:blab auspice/sars-like-cov*.json
nextstrain remote upload groups/blab auspice/sars-like-cov*.json
```

This draws on Docker's image syntax for inspiration. The nice parts of the first syntax are that a) it matches how groups are accessed on the web and b) supports local testing of the nextstrain.org server component by making the remote destination, e.g. `localhost:4000/groups/blab`. The above also aren't all mutually exclusive if we want to support shorter aliases for the longer, more general form.

#89 is a strict prerequisite. Changes to nextstrain.org's web API (what I sometimes call Charon v2) would also be really nice so that the CLI can talk to it and be unaware of AWS implementation details. (I also see such a Charon v2 API as providing a foundation for better static hosting of Auspice.) However, interim solutions exist where the CLI maps remotes for Nextstrain Groups to S3 buckets and talks to S3 behinds the scenes using temporary credentials granted by Cognito.

trvrb:
This is a really great direction. I also prefer (1) from an imagined user perspective. This cuts out redirection and makes it clear where the files are going.

jameshadfield:
Looks great! I also agree that syntax 1 seems the best.

> we'd like to support using nextstrain.org credentials

[...]

PR #92, opened 2020-06-25
Login

### LINKED ISSUE #89: Add login and logout commands to support Nextstrain Groups users
On first use:

```
$ nextstrain login
Logging into nextstrain.org…
Username:
Password:
Logged in successfully!
Credentials saved to ~/.nextstrain/secrets.
```
or
```
Failed to log in: {error}
```

Afterwards:

```
$ nextstrain login
Logged into nextstrain.org as {username}.
Log out with `nextstrain logout`.

$ nextstrain logout
Credentials removed from ~/.nextstrain/secrets.
Logged out.

$ nextstrain login --check
Checking credentials by logging into nextstrain.org…
Logged in successfully!
```

The `nextstrain login` command would prompt for nextstrain.org credentials, verify them, and store them (or maybe some shorter-lived derived credentials?) locally for later use (e.g. by `nextstrain remote upload` in the future).

Credentials can be stored in _~/.nextstrain/secrets_ under a `[login]` section. While the [keyring](https://pypi.org/project/keyring/) package makes cross-platform secret storage easier, it comes with a bit of initial setup overhead that might make CLI installation and usage much harder.

kairstenfay:
Implement `nextstrain login` and `nextstrain logout` for nextstrain.org groups users.

The following commands and options now exist:
`nextstrain login`
`nextstrain login --check`
~`nextstrain login --delete`~
`nextstrain logout`

### Related issue(s)  
Resolves #89   
Related to #90 

### Remaining TODOs
- [x] why is `jose` not found in the Pipfile when running Travis CI jobs?
- [x] some code duplication w.r.t. fetching credentials and logging in on error between `login` and `login --check` functionality. I plan to resolve this after first round of feedback to see if I've generally got the right idea solving #89 here

### DIFF
--- a/nextstrain/cli/command/login.py
+++ b/nextstrain/cli/command/login.py
@@ -5,7 +5,7 @@ import getpass
 from pycognito import Cognito
 from typing import Dict
 from .. import config
-from ..errors import UserError
+from ..errors import AuthnError, UserError
 
 
 COGNITO_CLIENT_ID='2vmc93kj4fiul8uv40uqge93m5'
@@ -14,6 +14,17 @@ COGNITO_USER_POOL_ID='us-east-1_Cg5rcTged'
 
 def register_parser(subparser):
     parser = subparser.add_parser("login", help = "Login as a Nextstrain.org Groups user")
+
+    register_arguments(parser)
+
+    return parser
+
+
+def register_arguments(parser):
+    parser.add_argument(
+        "--no-prompt",
+        help    = "Log into nextstrain.org without prompting for credentials",
+        action  = 'store_true')
 
     return parser
 
@@ -27,6 +38,9 @@ def run(opts):
     }
 
     if not tokens['username']:
+        if opts.no_prompt:
+            raise AuthnError("No Nextstrain.org Groups credentials found.")
+
         login()
         return 0
 
@@ -35,6 +49,8 @@ def run(opts):
 
     except Exception as e:
         error_message = "Failed to log in with saved credentials: %s" % e
+        if opts.no_prompt:
+            raise AuthnError(error_message)
 
         print(error_message)
         print()
--- a/nextstrain/cli/command/login.py
+++ b/nextstrain/cli/command/login.py
@@ -21,7 +21,56 @@ def register_parser(subparser):
 def run(opts):
     print("Logging into nextstrain.org...")
 
-    login()
+    tokens = {
+        k : config.get('login', k, fallback=None, path=config.SECRETS)
+            for k in [ 'username' ,'access_token', 'id_token', 'refresh_token' ]
+    }
+
+    if not tokens['username']:
+        login()
+        return 0
+
+    try:
+        authenticate_without_password(tokens)
+
+    except Exception as e:
+        error_message = "Failed to log in with saved credentials: %s" % e
+
+        print(error_message)
+        print()
+        login()
+        return 0
+
+    print("Logged into nextstrain.org as %s" % tokens['username'])
+    print("Log out with `nextstrain logout`")
+
+
+def authenticate_without_password(tokens: Dict[str, str]):
+    """
+    Authenticates a Nextstrain.org Groups user with the provided *tokens*.
+    """
+    cognito = Cognito(COGNITO_USER_POOL_ID, COGNITO_CLIENT_ID,
+        username=tokens['username'],
+        id_token=tokens['id_token'],
+        refresh_token=tokens['refresh_token'],
+        access_token=tokens['access_token'])
+
+    # The Cognito.authenticate() method, which performs token verification,
+    # requires a password. We do not want to store passwords locally or prompt
+    # a Nextstrain.org Groups user for a password each time they login
+    # (otherwise, what's the point of storing credentials locally?).
+    #
+    # So, as a workaround for authenticating AWS SRP tokens, utilize the
+    # undocumented Cognito.verify_token() method which will throw an error if
+    # the given token is invalid.
+    cognito.verify_token(tokens['id_token'], 'id_token', 'id')
+
+    # The Cognito.check_token() method verifies the access token, so there's no
+    # need to use the .verify_token() workaround here.
+    access_token_expired = cognito.check_token()
+    if access_token_expired:
+        print("Replacing expired tokens...")
+        save_credentials(cognito, config.SECRETS)
 
 
 def login():
--- a/nextstrain/cli/command/login.py
+++ b/nextstrain/cli/command/login.py
@@ -1,0 +1,67 @@
+"""
+Logs in a Nextstrain.org Groups user, storing their credentials locally.
+"""
+import getpass
+from pycognito import Cognito
+from typing import Dict
+from .. import config
+from ..errors import UserError
+
+
+COGNITO_CLIENT_ID='2vmc93kj4fiul8uv40uqge93m5'
+COGNITO_USER_POOL_ID='us-east-1_Cg5rcTged'
+
+
+def register_parser(subparser):
+    parser = subparser.add_parser("login", help = "Login as a Nextstrain.org Groups user")
+
+    return parser
+
+
+def run(opts):
+    print("Logging into nextstrain.org...")
+
+    login()
+
+
+def login():
+    """
+    With user input, logs a Nextstrain Groups user into nextstrain.org.
+
+    Failed login attempts are printed to stdout and result in a non-zero exit
+    code.
+
+    Saves SRP credentials to a local secrets path.
+    """
+    username = input('Username: ')
+    password = getpass.getpass()
+
+    try:
+        cognito = Cognito(COGNITO_USER_POOL_ID, COGNITO_CLIENT_ID, username=username)
+        cognito.authenticate(password=password)
+
+    except Exception as e:
+        raise UserError("Failed to login: %s" % e)
+
+    print("Logged in successfully!")
+
+    save_credentials(cognito, config.SECRETS)
+
+
+def save_credentials(c: Cognito, path):
+    """
+    Saves SRP tokens from the given *cognito* instance to a local file at *path*
+    under a 'login' header.
+    """
+    tokens = {
+        'username'      : c.username,
+        'id_token'      : c.id_token,
+        'refresh_token' : c.refresh_token,
+        'access_token'  : c.access_token,
+        'token_type'    : c.token_type
+    }
+
+    for key in tokens:
+        config.set('login', key, tokens[key], path)
+
+    print("Credentials saved to %s" % path)
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -129,6 +129,48 @@
             ],
             "version": "==2020.12.5"
         },
+        "cffi": {
+            "hashes": [
+                "sha256:005a36f41773e148deac64b08f233873a4d0c18b053d37da83f6af4d9087b813",
+                "sha256:0857f0ae312d855239a55c81ef453ee8fd24136eaba8e87a2eceba644c0d4c06",
+                "sha256:1071534bbbf8cbb31b498d5d9db0f274f2f7a865adca4ae429e147ba40f73dea",
+                "sha256:158d0d15119b4b7ff6b926536763dc0714313aa59e320ddf787502c70c4d4bee",
+                "sha256:1f436816fc868b098b0d63b8920de7d208c90a67212546d02f84fe78a9c26396",
+                "sha256:2894f2df484ff56d717bead0a5c2abb6b9d2bf26d6960c4604d5c48bbc30ee73",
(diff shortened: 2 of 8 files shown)


kairstenfay:
Looking through some `nextstrain remote` code now, and it looks like I can replace all instances in this P.R. of `sys.exit(#)` with `return #`.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+            print("No credentials found to delete!")
+
+        sys.exit(0)
+
+    if opts.check:
+        print("Checking credentials by logging into nextstrain.org...")
+        try:
+            config.read(secrets_path)
+            username = config.get('login', 'username')
+            authn_result = ast.literal_eval(config.get('login', 'authenticationresult'))
```
tsibley, line 58: 🤨 This makes me raise an eyebrow as unnecessary complexity/use of the AST parser. Instead of storing a stringified dict, why not store the tokens under individual keys (e.g. `config.get("id_token")`, `config.get("refresh_token")`, …)?

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # If the requested credentials are not found in a 'login' section in the
+        # specified secrets file, generate them, and save them locally
+        except (FileNotFoundError, configparser.NoSectionError, configparser.NoOptionError):
+            login()
+            sys.exit(0)
+
+        u = Cognito(os.environ['AWS_COGNITO_USER_POOL_ID'],
+                '2vmc93kj4fiul8uv40uqge93m5',
```
tsibley, line 67: If we declare this opaque magic value as `COGNITO_CLIENT_ID` at the top of the file, then it's self-documenting. :-)

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+            username = config.get('login', 'username')
+            authn_result = ast.literal_eval(config.get('login', 'authenticationresult'))
+
+        # If the requested credentials are not found in a 'login' section in the
+        # specified secrets file, generate them, and save them locally
+        except (FileNotFoundError, configparser.NoSectionError, configparser.NoOptionError):
+            login()
+            sys.exit(0)
+
+        u = Cognito(os.environ['AWS_COGNITO_USER_POOL_ID'],
```
tsibley, line 66: The user pool id should be hardcoded in source, not come from the environment. Remember that the CLI runs on many people's computers, and no one will have or should have to have a specific environment variable set. (If necessary, we could support _overriding_ it from the environment, for things like testing/dev, but I don't think that's necessary here.)

Best to declare the hardcoded value as `COGNITO_USER_POOL_ID` at the top of the file.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+                '2vmc93kj4fiul8uv40uqge93m5',
+                username=username,
+                id_token=authn_result['IdToken'],
+                refresh_token=authn_result['RefreshToken'],
+                access_token=authn_result['AccessToken'])
+
+        try:
+            u.verify_token(authn_result['IdToken'], 'id_token', 'id')
+            u.verify_token(authn_result['AccessToken'], 'access_token', 'access')
```
tsibley, line 75: I worry slightly about using this undocumented part of the pycognito API.

nextstrain/cli/command/login.py
tsibley: Actually, what's storing `authenticationresult`? I don't see if referenced anywhere else.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+            sys.exit(0)
+
+        access_token_expired = u.check_token()
+
+        if not access_token_expired:
+            print("Logged in succesfully!")
+
+        else:
+            print("Replacing expired access token...")
+            tokens['AccessToken'] = u.access_token
```
tsibley, line 90: Ah, I think we need to store all the tokens if we're going to use any of them.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
+"""
+Logs in a Nextstrain.org groups user, storing their credentials locally.
```
tsibley, line 2: Minor thing, but I've been capitalizing Nextstrain Groups (and even just when "Groups" in passing) as a way to indicate that it's a capital-T Thing and distinguish it from the more general meaning of "group".

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+
+    try:
+        client = boto3.client('cognito-idp')
+
+        # Use AWSSRP utility class to easily save credentials
+        aws = AWSSRP(
+                username=username,
+                password=password,
+                pool_id=os.environ['AWS_COGNITO_USER_POOL_ID'],
+                client_id='2vmc93kj4fiul8uv40uqge93m5',
```
tsibley, line 129: Ditto with magic values here as above.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+                client=client
+            )
+        tokens = aws.authenticate_user()
+
+    except Exception as e:
+        print(f"Failed to login: {e}")
+        sys.exit(1)
```
tsibley, line 136: Better to throw a [`NextstrainUserError`](https://github.com/nextstrain/cli/blob/972803cd683ba7543d836f5d51a158ebe8034bff/nextstrain/cli/errors.py) instead of calling `sys.exit` directly so that deeper parts of the code aren't dead ends but pop the stack all the way back up with the exception. This allows a cleaner design with [a single point](https://github.com/nextstrain/cli/blob/972803cd683ba7543d836f5d51a158ebe8034bff/nextstrain/cli/__init__.py#L34-L39) to control expected exit statuses and error handling. (You may be familiar with the same concept people use in the React world, called "error boundaries" there.)

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Use AWSSRP utility class to easily save credentials
+        aws = AWSSRP(
+                username=username,
+                password=password,
+                pool_id=os.environ['AWS_COGNITO_USER_POOL_ID'],
+                client_id='2vmc93kj4fiul8uv40uqge93m5',
+                client=client
+            )
+        tokens = aws.authenticate_user()
```
tsibley, line 132: I think it'd be a little bit nicer to use pycognito's API for the light abstraction it provides over the direct boto3 client access. For example, its `authenticate()` method includes token verification, which is missing here.

.nextstrain/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
tsibley, line 1: This directory shouldn't exist (see my other comment), so this gitignore should go away.

Pipfile
```diff
@@ -8,6 +8,8 @@ verify_ssl = true
 [packages]
 nextstrain-cli = {editable = true,extras = ["dev"],path = "."}
 sphinx-argparse = {file = "https://github.com/alex-rudakov/sphinx-argparse/archive/b3b649743d4f8854349a2416ccdd770f41739c35.tar.gz"}
+pycognito = "*"
+jose = "*"
```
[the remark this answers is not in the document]
kairstenfay: Ah, I missed that `nextstrain-cli` is a dependency here. That would have been a bigger clue.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+def register_arguments(parser):
+    parser.add_argument(
+        "--check",
+        help    = "Check credentials by logging into nextstrain.org",
+        action  = 'store_true')
+
+    parser.add_argument(
+        "--delete",
+        help    = "Deletes the secrets file containing authentication credentials",
+        action  = 'store_true')
```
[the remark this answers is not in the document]
kairstenfay: Great, I'll delete it.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,153 @@
...
+import boto3
+import getpass
+import configparser
+from pathlib import Path
+from pycognito import Cognito
+from pycognito.aws_srp import AWSSRP
+from typing import Dict
+
+base = Path(__file__).resolve().parent.parent.parent.parent
+secrets_path = base / '.nextstrain/secrets'
```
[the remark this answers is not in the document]
kairstenfay: Ah, for some reason I was confusing the tilde in ~/.nextstrain/config for the base of this repo. This makes sense to me now.

nextstrain/cli/command/login.py
kairstenfay: Gotcha, I wasn't sure if it was a secret or not, so I erred on the safer side.

nextstrain/cli/command/login.py
kairstenfay: `authenticationresult` was one of they keys in the dict returned by the AWSSRP utility class that I in turn saved to the config. It will probably not be relevant after switching to using only the Cognito utility class.

nextstrain/cli/command/login.py
tsibley: :+1: That's a good instinct! In our case, it's not a secret. (Other Cognito use cases may treat it as a semi-secret.)

For reference, you can also see what other Cognito values we treat as public or not for nextstrain.org authentication:

https://github.com/nextstrain/nextstrain.org/blob/497d62e30983042621436a13af6a6df21d48b166/authn.js#L27-L35

nextstrain/cli/command/login.py
kairstenfay: I wasn't sure how else to verify that the provided tokens are correct, i.e. not corrupted somehow. There is no obvious, documented method of Cognito that provides this functionality, since `authenticate()` requires a password, which is not what I need when I do a `--check`.

kairstenfay: Perhaps I'm mistaken, though, and prompting a logged-in user for their password when they run login with `--check` is exactly what I should be doing. I am assuming that I don't want to store their pwd in plain text in ~/.nextstrain/secrets

kairstenfay:
@tsibley I believe I've incorporated all of your feedback but would still appreciate input on `--check` (see the unresolved convo about `verify_token()` above.

nextstrain/cli/command/login.py
kairstenfay: Thanks for pointing out that file!

nextstrain/cli/command/login.py
kairstenfay: I think I was already storing them? But let me know if I missed something here.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,127 @@
...
+
+    print("Logged in successfully!")
+    tokens = {
+        'id_token'          : u.id_token,
+        'refresh_token'     : u.refresh_token,
+        'access_token'      : u.access_token,
+        # TODO the pycognito docs say the Cognito instance should have the
+        # following "expires_*" attributes, but it doesn't.
+        # 'expires_in'        : u.expires_in,
+        # 'expires_datetime'  : u.expires_datetime,
```
kairstenfay, line 113: I'm not sure what to do here, as it appears this is in direct contradiction with pycognito's docs, but I'll take a closer look again tomorrow.

nextstrain/cli/command/login.py
```diff
@@ -0,0 +1,127 @@
...
+        u = Cognito(COGNITO_USER_POOL_ID, COGNITO_CLIENT_ID, username=username,
+                id_token=tokens['id_token'],
+                refresh_token=tokens['refresh_token'],
+                access_token=tokens['access_token'])
+
+        try:
+            u.verify_token(tokens['id_token'], 'id_token', 'id')
+            u.verify_token(tokens['access_token'], 'access_token', 'access')
+
+        except (jose.exceptions.JWSError, jose.exceptions.JWTError):
```
tsibley, line 56: I think we can avoid knowing the implementation detail that `pycognito` uses `jose` and instead catch `pycognito.TokenVerificationException`. This also means we don't have to explicitly dep or import `jose`.

kairstenfay:
Ready for re-review.

tsibley:
Rebased this onto latest master last night.

Also wanted to note that I audited pycognito 2021.02.0 to confirm that

1. The source and wheel packages on PyPI match each other.
2. The PyPI packages match [tagged commit](https://github.com/pvizeli/pycognito/commit/cfe08edd1690beca32c48ade1928660ea76c8855) on GitHub.
3. There's no obvious funny-business in that version of the code. (Although there is a port of some finicky cryptography that could have bugs, but reviewing that is beyond my qualifications.)

tsibley:
In local testing with this, I ran into issues with old tokens, and started looking into smoothing them out. In the process, I believe I found some weaknesses in pycognito's handling of tokens, so I'll be submitting PRs upstream too.

tsibley:
See https://github.com/pvizeli/pycognito/pull/38.

PR #101, opened 2020-09-29
Upload individual output files from AWS Batch job

ivan-aksamentov:

This adds uploading of the individual files in the AWS Batch job working directory.

This is handy for situations when only one file needs to be downloaded, without downloading of the whole archive, which can be several gigabytes in size. Proposed by @emmahodcroft 

Note that this does not remove the upload of the large zip. Mostly because I simply don't know how it is used and whether there are tools that rely on its presence.

Implementation is a heavy copy-pate of the existing `upload_workdir()` function:
https://github.com/nextstrain/cli/blob/5bb2ccd4ec58c222015acd82d7fdc7fd4feca3cc/nextstrain/cli/runner/aws_batch/s3.py#L63-L71
with the loop logic inverted: instead of adding each file to the zip, we create a new zip for each file:
https://github.com/nextstrain/cli/blob/5bb2ccd4ec58c222015acd82d7fdc7fd4feca3cc/nextstrain/cli/runner/aws_batch/s3.py#L97-L105

### Related issue(s)  

N/A

### Testing
I guess we need to try and trigger an AWS Batch-hosted build and see how it goes :)

### Thank you for contributing to Nextstrain!
My pleasure :)

### DIFF
--- a/nextstrain/cli/runner/aws_batch/__init__.py
+++ b/nextstrain/cli/runner/aws_batch/__init__.py
@@ -123,6 +123,10 @@ def run(opts, argv, working_volume = None, extra_env = {}, cpus: int = None, mem
 
         print("uploaded:", s3.object_url(remote_workdir))
 
+        remote_files = s3.upload_workdir_individually(local_workdir, bucket, run_id)
+
+        for remote_file in remote_files:
+            print("uploaded:", s3.object_url(remote_file))
 
         # Submit job.
         print_stage("Submitting job")
--- a/nextstrain/cli/runner/aws_batch/s3.py
+++ b/nextstrain/cli/runner/aws_batch/s3.py
@@ -7,7 +7,7 @@ import boto3
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from calendar import timegm
-from os import utime
+from os import utime, path as os_path
 from pathlib import Path
 from tempfile import TemporaryFile
 from time import struct_time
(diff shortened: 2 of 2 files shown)


ivan-aksamentov:
The build seems to be currently failing on mypy step with an error:

```
Format strings are only supported in Python 3.6 and greater
```

Looks like we target pre-3.6 as well, so I'll downgrade the syntax.

ivan-aksamentov:
Internal Slack discussion:
https://bedfordlab.slack.com/archives/C0H7C3F0B/p1601394339002700

tsibley:
Recording the below here for posterity and future reference.

About a month ago this PR briefly came up in Slack, where the reason for its existence was:

> _[The ncov build shepherds]_ often need to download only some small files from the results of the job, but are currently forced to download everything (the whole zip).  This PR adds zipping ad uploading every file separately.

I took a quick look at it and [concluded](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1604694235176600?thread_ts=1604662851.145800&cid=CTZKJC7PZ):

> I think maybe that PR isn't doing what you want? It's uploading files from the local computer to S3, not uploading files from the Batch job to S3.

ivan-aksamentov:
Yes, this is a complete bogus, sorry. :flushed: 
I will open an issue instead.

tsibley:
No worries, thanks!

PR #105, opened 2021-01-25
add docs build env specifying theme version

eharkins:
following pattern of other subprojects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.pyc
 __pycache__/
-environment*
 
 # Package build dirs
 /build/
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -1,0 +1,12 @@
+name: cli-docs
+channels:
+  - defaults
+dependencies:
+  - make
+  - recommonmark
+  - sphinx
+  - pip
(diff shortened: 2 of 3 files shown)


tsibley (changes requested):
Ah, I think you might have missed the existing `.readthedocs.yml` file (note the leading dot) and corresponding `.readthedocs-requirements.txt` file, which already specifies the theme version. Dev requirements like this flow from `setup.py` → Pipenv (`Pipfile` + `Pipfile.lock`), and then for RTD from Pipenv → `.readthedocs-requirements.txt`, which is used in a `pip install` by `.readthedocs.yml`.

I would like to ditch Pipenv (which these days causes more headache than help compared to 2y ago) sooner than later, but I don't want to add yet another dep management tool (Conda) to the mix right now.

eharkins:
My mistake! I think I was mostly trying to address a perceived issue where I wasn't getting the CLI docs with the most recent nextstrain sphinx theme version, but that must have been when building them locally due to my misunderstanding of the build environment configuration, since it seems to be up to date at https://docs.nextstrain.org/projects/cli/en/latest/

PR #115, opened 2021-03-08
Login, take two

tsibley:
Login and logout commands for authenticating with nextstrain.org. This PR lays the foundation for other commands to make authenticated requests to nextstrain.org's APIs, however, currently no commands make such requests.

All the details are in the commit messages. The best overview is 296a82aaf395adba3a13899d8e9651da4fcdc53f.

Supersedes https://github.com/nextstrain/cli/pull/92, where a lot of design discussion took place.

Todo (but not blocking initial review):

- [x] Support initial login credentials from the environment (`NEXTSTRAIN_USERNAME`, `NEXTSTRAIN_PASSWORD`) instead of prompting, to support automated environments like CI builds.


trvrb:
Thanks Tom! This is amazing. I like this from a user perspective and confirmed it was as intended with my account.

The general strategy seems smart, but I don't have have any expert insight into how to securely manage credentials.

huddlej (approved):
This "worked" for me (as someone who doesn't have a Nextstrain account) in that all the new commands behaved as expected. The code all generally looks good, too, although like Trevor I can't comment on the credential management aspects.

tsibley:
@trvrb @huddlej Thanks for taking a look! I'm mostly looking for feedback on the UI/UX and overall organization/approach of the code, so glad that seems good. The credentials management itself is comparable to other tools, and I think fine.

I did just repush to remove the group read permissions from the secrets file (leaving it `u=rw,go=`), as upon further consideration other tools (like ssh) do this as well. The rebase also resolves some conflicts subsequently introduced on master.

My thought is to merge this now (once CI passes again), even though nothing is making use of it yet. Some folks may notice it and wonder about it, but I think that's fine.

jameshadfield:
> My thought is to merge this now (once CI passes again), even though nothing is making use of it yet. Some folks may notice it and wonder about it, but I think that's fine.

Seems good to me. I'll try out the commands (UI/UX) now if you can hold off for a bit longer...

tsibley:
@jameshadfield Yes! There's no rush (other than the nice feeling of merging a bunch of work :-).

jameshadfield:
The login/logout/whoami commands worked as expected 💯 

### UI / UX comments

* There's probably going to be some confusion from people who have been using `nextstrain remote upload...` for groups, which use IAM credentials from the environment not cognito usernames, who try to use these credentials as logins. There are few enough groups that this could be managed on a case-by-case basis, as I imagine checking if a username is actually an IAM access key is a bit too much. 
* I've gotten in the habit of running `--help-all` instead of `--help` (for the CLI, i'm not aware of other software which uses this split), so it'd be nice if that would work for commands like `nextstrain login` etc, even if the output was identical to `--help`. 


### Install troubles

Perhaps unrelated to this PR as I was upgrading from an older version of the CLI on this machine, but when re-installing the cli (to ensure deps up to date) I got the following warning:
```
$ pip install -e .[dev]
...
botocore 1.19.52 has requirement urllib3<1.27,>=1.25.4; python_version != "3.4", but you'll have urllib3 1.24 which is incompatible.
```

Which then resulted in a broken CLI:
```
$ nextstrain --help
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/pkg_resources/__init__.py", line 578, in _build_master
    ws.require(__requires__)
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/pkg_resources/__init__.py", line 895, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (urllib3 1.24 (/Users/naboo/miniconda3/envs/nextstrain-cli/lib/python3.7/site-packages), Requirement.parse('urllib3<1.27,>=1.25.4; python_version != "3.4"'), {'botocore'})
```

The piplock pins this to `urllib3==1.26.3` but since I don't use pipenv I ended up added the following to `setup.py` to get things working

```py
        "requests == 2.25.1", # previously unpinned which gave me v2.20 which demands urllib3 < 1.24
        "urllib3 == 1.26.3", # same as in pip.lock
 ```

tsibley:
> There's probably going to be some confusion from people who have been using `nextstrain remote upload...` for groups, which use IAM credentials from the environment not cognito usernames, who try to use these credentials as logins. There are few enough groups that this could be managed on a case-by-case basis, as I imagine checking if a username is actually an IAM access key is a bit too much.

Agreed there may be confusion here for the current "early adopters" of Groups.

Agreed it's likely too much to call the AWS STS [`GetAccessKeyInfo` method](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetAccessKeyInfo.html) for each username given (and doing so would presume available AWS credentials), but it'd be plausible to add a heuristic based on the given username matching something like `/^A[KS]IA[A-Za-z0-9]{12,}$/`.

That said, I'm not sure it's worth the extra effort here, for what would probably be a one-time confusion for at max a dozen or two people. It'd be fairly easy to add, I think, so maybe even a little confusion-avoided is worth it. Thoughts?

> I've gotten in the habit of running `--help-all` instead of `--help` (for the CLI, i'm not aware of other software which uses this split), so it'd be nice if that would work for commands like `nextstrain login` etc, even if the output was identical to `--help`.

Mmm, that's a reasonable request. Good idea and doable. I've opened https://github.com/nextstrain/cli/issues/117.

(The `--help` / `--help-all` split is one I've encountered a few times over the years (sometimes with a naming variation), though I don't have examples on-hand to point to. I would like to keep it, as it simplifies the interface for most users.)

> Perhaps unrelated to this PR as I was upgrading from an older version of the CLI on this machine, but when re-installing the cli (to ensure deps up to date) I got the following warning:

Can you provide:

1. The older version of the CLI you were upgrading from?
2. The version of pip in your Conda env?

The transitive dep resolution around `botocore` is more than a bit fraught because of some of the choices of our direct dependencies, but I haven't seen this particular issue before and can't reproduce locally with some basic attempts.

tsibley:
It seems as if pip didn't want to upgrade an existing package despite a req for a newer version, which is confusing. I wonder if you needed to pass the `--upgrade` flag to `pip install` to allow it to upgrade transitive deps? (While I know how `--upgrade` affects direct packages, I'm not sure of the exact impact on transitive deps.)

jameshadfield:
Re: confusion of early adopters. I think this can be overcome with clear documentation, rather than code changes. When we publicise this, presumably when the first functions start to take advantage (`nextstrain remote`?), we can make a special note of this. If some people make the mistake, then we can deal with it on a case-by-case basis. 

Re: install issues. This was in a rather out-of-date conda environment, and I don't mean for it to block merging of this PR at all. However if we run into these things then I imagine others will too! I was using a conda environment with python 3.7, pip 18.1, nextstrain-cli v1.x (don't remember which exactly, perhaps 1.16).

tsibley:
> I think this can be overcome with clear documentation, rather than code changes.

:+1: 

> However if we run into these things then I imagine others will too! I was using a conda environment with python 3.7, pip 18.1, nextstrain-cli v1.x (don't remember which exactly, perhaps 1.16).

Yeah, quite possibly! Thanks for the details; I think they may be helpful in troubleshooting the issue if we see it again. (I'm not going to try to reproduce right now, though.)

PR #134, opened 2021-11-03
Adjust deps to avoid incompatible versions of s3fs and aiobotocore

### LINKED ISSUE #133: [batch] Unable to submit builds when installed with latest aiobotocore
**Current Behavior**  
Throws an `AttributeError: module 'aiobotocore' has no attribute 'AioSession'` during upload of the workdir to S3, causing cascading failures.

[First observed](https://bedfordlab.slack.com/archives/CTZKJC7PZ/p1635927498059100) in our ncov-ingest runs ([example](https://github.com/nextstrain/ncov-ingest/runs/4088163851?check_suite_focus=true#step:3:15)). [Workaround for that repo](https://github.com/nextstrain/ncov-ingest/pull/227).

https://github.com/nextstrain/ncov-ingest/pull/227#issuecomment-959776198 has the best summary of what's happening with pip's dep resolution and the dep declarations of nextstrain-cli/s3fs/fsspec/aiobotocore.

**Expected behavior**  
It works.

**How to reproduce**  
```
mkdir /tmp/venv
python3 -m venv /tmp/venv
source /tmp/venv/bin/activate
pip install --upgrade pip setuptools wheel
pip install nextstrain-cli
nextstrain build --aws-batch zika-tutorial
```

**Possible solution**  
TKTK.

tsibley:
Affects fresh installs of nextstrain-cli with the latest aiobotocore
version, 2.0.0.

The latest aiobotocore, 2.0.0, does not fulfill the dep declarations of
the latest s3fs, 2021.10.1.  As a result, pip fulfills our s3fs dep with
2021.7.0 instead, which while declared to be compatible with
aiobotocore>=1.4.1 is not actually compatible in practice.  The result
is a runtime AttributeError on AioSession during upload to S3 in
`nextstrain build --aws-batch`.

When our previous deps on s3fs and aiobotocore[boto3] were introduced,
s3fs did not yet support its own [boto3] extra.  It does now, which
means we can let it declare the aiobotocore[boto3] dep itself and not
end up with conflicting dep declarations.  This results in fresh
installs using an older version of aiobotocore (1.4.x), but that's what
we want since it works.  Good to get the transitive dep out of our
declarations too.

Still a mess.  Dependencies are hard.

Resolves #133.


joverlee521:
Does this change also require updating `.readthedocs-requirements.txt`?

tsibley:
Yeah, doing so locally but got distracted by this meeting. ;-)

tsibley:
@joverlee521 Oh man, that was a journey I didn't want to take, but this PR should now be passing all CI. orz

joverlee521 (approved):
Thank you for working through this @tsibley 
Pipenv really hasn't helped with our dependency hell (in SFS or here) 🙈

PR #140, opened 2022-01-20
Docs and --help integration

tsibley:
Organizational and process improvements to the RTD docs and `--help` output and better integration between the two.

Original impetus for this was firmer ground for writing docs and help for the `nextstrain remote` family of commands' second supported remote (nextstrain.org) in addition to the existing S3 remote. In the process made some organizational/table of contents changes as well as I thought more about how to best present this information.

Details in the commit messages, as usual.


jameshadfield:
Built docs here: https://nextstrain--140.org.readthedocs.build/projects/cli/en/140/

doc/index.rst
```diff
@@ -19,14 +19,38 @@ across computing environments such as `Docker <https://docker.com>`__, `Conda
     :doc:`docs:tutorials/running-a-workflow` tutorial first and then come back
     here.
 
+
```
jameshadfield, line 22: I'm not sure a TOC is necessary on the splash page, but if so I'd recommend the "Big Picture" section to be lifted above the TOC as it's getting buried at the moment.

jameshadfield (changes requested):
Looks good Tom - I only reviewed the docs reorganisation, not all the sphinx / parsing work. Would you mind explaining what's going on there in a sentence or two? 

I'd push back on commit babf294ee3968ce86dfc7b747c9d683c6f012754 ("We can't stuff all the information we'd want into --help..."). As we are in the process of modifying how authn works, it'd be nice to keep surfacing via `--help` that AWS keys are what's used. Additionally, I'd suggest adding a brief sentence to login/logout that they are _not_ currently used for authn (unless, of course, they are and I've missed that feature release!).

doc/index.rst
tsibley: Yeah. When I was working on these changes, I thought about moving the "Big Picture" diagram up because it's definitely buried now. But I realized it's a bit out of date, so I didn't want to make it more prominent yet. The doc home page here could definitely use some more improvements such as that and more consideration of what's presented (e.g. TOC or no TOC). Agreed that a full TOC may not be necessary, but also the current "Commands" TOC on the home page is used as the [target of a redirect](https://github.com/nextstrain/cli/blob/75afe1227449603db8e4a0799a227c88be5e990d/doc/redirects.yaml#L11-L17) for the old `doc/commands/index.rst` page. Anyway, I wanted to postpone a full home page rework for now.

tsibley:
> Looks good Tom - I only reviewed the docs reorganisation, not all the sphinx / parsing work. Would you mind explaining what's going on there in a sentence or two? I got lost with all the rst / plain text stuff.

Command descriptions appear in both `--help` output and online doc pages. Previously, `--help` displayed the raw description as-is while the online doc pages rendered the description as rST → HTML (using Sphinx + Docutils). Descriptions had to balance between using rST syntax to improve the online pages and still keeping the `--help` output readable. Now, `--help` renders the description as rST → plain text (using Docutils, sans Sphinx).

(I've added this explanation to f9dfe4b.)

> I'd push back on commit [babf294](https://github.com/nextstrain/cli/commit/babf294ee3968ce86dfc7b747c9d683c6f012754) ("We can't stuff all the information we'd want into --help..."). As we are in the process of modifying how authn works, it'd be nice to keep surfacing via `--help` that AWS keys are what's used. Additionally, I'd suggest adding a brief sentence to login/logout that they are _not_ currently used for authn (unless, of course, they are and I've missed that feature release!).

Hmm. The support for nextstrain.org as a remote is forthcoming and builds upon this PR's branch. Maybe it makes more sense to leave the `nextstrain remote --help` as-is for now and make the changes in the next PR along with the nextstrain.org remote docs. I think I'll do that.

doc/index.rst
```diff
@@ -2,32 +2,55 @@
...
-as `Docker <https://docker.com>`__, `Conda
+.. hint::
+   This is reference documentation for the Nextstrain CLI (command-line
+   interface).  If you're just getting started with Nextstrain, please see
+   :doc:`our general documentation <docs:index>` instead.
+
+The Nextstrain CLI is a program called ``nextstrain``.  It aims to provide a
+consistent way to run and visualize pathogen builds and access Nextstrain
+components like :doc:`Augur <augur:index>` and :doc:`Auspice <auspice:index>`
+across computing environments such as `Docker <https://docker.com>`__, `Conda
```
victorlin, line 13: [runtime](https://docs.nextstrain.org/en/latest/reference/glossary.html#term-runtime)?

tsibley: Yes! …but I held off in this change/PR because other places in the CLI docs still using "computing environment" and doing that switch wasn't in the scope of my changes here. I definitely want the switch to "runtime" to happen, but want it to happen all at once to keep the docs internally consistent.

victorlin (approved):

PR #141, opened 2022-01-25
Nextstrain.org remote

tsibley:
Builds upon #140 to add nextstrain.org support to the `nextstrain remote` family of commands.

### Kicking the tires

Works well in my extensive local testing against https://github.com/nextstrain/nextstrain.org/pull/452. To do so yourself, run the nextstrain.org server from that PR:

```
# in checkout of nextstrain/nextstrain.org repo
git switch restful-api-rw
npm run build
node server.js --verbose
```

and then in another shell, before running `nextstrain remote` commands from this PR, make sure to:

```
export NEXTSTRAIN_DOT_ORG=http://localhost:5000
```

As a confirmation, you should see `localhost:5000` in the output of `nextstrain remote ls nextstrain.org`, e.g.:

```
$ nextstrain remote ls nextstrain.org
http://localhost:5000/dengue/denv1
http://localhost:5000/dengue/denv2
http://localhost:5000/dengue/denv3
http://localhost:5000/dengue/denv4
http://localhost:5000/dengue/all
http://localhost:5000/ebola
…
```


victorlin:
Diff between this and #140 for easier review: https://github.com/nextstrain/cli/compare/trs/docs-and-help...trs/remote/nextstrain.org

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,697 @@
...
+    List the :class:`Resource`(s) available at *path*.
+
+    If *recursively* is false (the default), then only an exact match of
+    *path* is returned, if any.  If *recursively* is true, then all resources
+    at or beneath *path* are returned.
+
+    If *http* is not provided, a new :class:`requests.Session` will be created.
+
+    If *authorization* is not provided, :func:`auth_headers` will be called to
+    obtain it.
```
victorlin, line 371: Is there a specific docstring format being followed here? Looks like restructured text but I don't see this being rendered on any page in the preview.

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,697 @@
...
+                    stream = True)
+
+                with response:
+                    # Skip/ignore missing sidecars
+                    if response.status_code == 404 and not subresource.primary:
+                        continue
+
+                    # Check for bad response
+                    raise_for_status(response)
+                    assert content_media_type(response) == subresource.media_type
```
victorlin, line 298: I ran `nextstrain remote download https://nextstrain.org/ebola` locally and got this error:

```
% nextstrain remote download https://nextstrain.org/ebola                    
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/__init__.py", line 33, in run
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/command/remote/download.py", line 89, in run
    for remote_file, local_file in downloads:
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/remote/nextstrain_dot_org.py", line 298, in download
    assert content_media_type(response) == subresource.media_type
AssertionError


An error occurred (see above) that likely indicates a bug in the
Nextstrain CLI.

To report this, please open a new issue and include the error above:
    <https://github.com/nextstrain/cli/issues/new/choose>
```

It looks like `content_media_type(response) = 'text/html'` and `subresource.media_type = 'application/vnd.nextstrain.dataset.main+json'`. Am I using this properly?

nextstrain/cli/remote/nextstrain_dot_org.py
tsibley: rST, yes, but the docstrings are somewhat loose and vary with when they were written. The internal interfaces aren't included in the Sphinx docs, as I'm not really sure who the audience would be for that. Augur [includes its internal interfaces](https://docs.nextstrain.org/projects/augur/en/stable/api/api.html) (generated from docstrings), but it actually has an expectation of being used as a library unlike the Nextstrain CLI.

nextstrain/cli/remote/nextstrain_dot_org.py
tsibley: Hmm. Did you start a local nextstrain.org server from the `restful-api` or `restful-api-rw` branch and set `NEXTSTRAIN_DOT_ORG=http://localhost:5000` before running `nextstrain remote`?

victorlin: Yes, using `restful-api-rw` branch. `nextstrain remote ls` works fine:

```
% echo "$NEXTSTRAIN_DOT_ORG"                             
http://localhost:5000
% nextstrain remote download https://nextstrain.org/ebola
'text/html'
'application/vnd.nextstrain.dataset.main+json'
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/__init__.py", line 33, in run
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/command/remote/download.py", line 89, in run
    for remote_file, local_file in downloads:
  File "/Users/vlin/Dropbox/repos/cli/nextstrain/cli/remote/nextstrain_dot_org.py", line 300, in download
    assert content_media_type(response) == subresource.media_type
AssertionError


An error occurred (see above) that likely indicates a bug in the
Nextstrain CLI.

To report this, please open a new issue and include the error above:
    <https://github.com/nextstrain/cli/issues/new/choose>

% nextstrain remote ls https://nextstrain.org/     
https://nextstrain.org/dengue/denv1
https://nextstrain.org/dengue/denv2
https://nextstrain.org/dengue/denv3
https://nextstrain.org/dengue/denv4
...
```

tsibley: `https://nextstrain.org/ebola` works just fine for me. I suspect you didn't export `NEXTSTRAIN_DOT_ORG`, so it's visible to your shell (as the `echo` shows) but not to any processes your shell launches.

Ah, and actually your `nextstrain remote ls` confirms this: the output would show the overridden URL, e.g. `http://localhost:5000/dengue/denv1` and so on.

victorlin: Whoops that's it, works fine with `export`. False alarm!

tsibley: Oh good. :-) This also motivated me to update the instructions at the top to be a bit clearer.

victorlin (approved):
Looks good to me with some basic testing! Learned a few things while looking over this too 🙂

Let me know if there's anything specific you'd like me to test locally.

tsibley:
Thanks for giving it a whirl! Aside from "does it work?" testing, I think the next thing to focus on is how the command behaviour/UX feels. Does it feel right? Is there anything that feels off or that you reached for but didn't work/exist?

Slightly tangentially, one missing thing I'd like to add support for (and which I think @jameshadfield also wants) is specifying a constant filename prefix to local files when doing `download --recursively`. Current that mode only supports specifying a local directory, but I think I know how to extend that reasonably.

tsibley:
My plan is to wait to merge this until nextstrain/nextstrain.org#452 gets deployed, as otherwise it won't work as advertised.

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,656 @@
...
+                            #       a/b
+                            #       a/b/x
+                            #       a/b/y
+                            #       a/b/z
+                            #
+                            # then we need to include a little of the given
+                            # path in the local filenames, e.g. strip only "a",
+                            # leaving "b".  Otherwise, e.g. if a/b doesn't
+                            # exist itself only a/b/x, etc, then we can strip
+                            # "a/b" entirely.
```
joverlee521, line 317: Hmm, this behavior was very unexpected for me when I was testing the recursive download.

Flu datasets might be a very specific use case:
1. run `remote download -r groups/test/flu /tmp/testing` 
2. run `remote download -r groups/test/flu/seasonal /tmp/testing` 

I expected the second command to just overwrite the existing files within `/tmp/testing`. Instead I got two sets of the same files but with slightly different names because the prefixes were stripped.

joverlee521 (approved):
Very cool to see this work so well! I mostly left comments about the general "feel" of the commands, so nothing blocking. 

A couple general comments I have:
1. It would be nice to have a `--dry-run` option for these commands.
2. Should user credentials be renewed before running each command? 
Currently, if I remove myself from a Cognito group, I can still run commands for the group as long as I don't renew my local tokens.

nextstrain/cli/remote/__init__.py
```diff
@@ -30,25 +32,43 @@
...
 def parse_remote_path(path: str) -> Tuple[RemoteModule, ParseResult]:
     url = urlparse(path)
 
-    if url.scheme not in SUPPORTED_SCHEMES:
-        raise UserError(f"""
-            Unsupported remote scheme {url.scheme}://
+    if not url.scheme or url.scheme in {"https", "http"}:
+        if not url.scheme:
+            if url.path.startswith("groups/"):
+                # Special-case groups/… as a shortcut
```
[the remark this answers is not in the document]
tsibley: It's online, where you can see it in the preview build: https://nextstrain--141.org.readthedocs.build/projects/cli/en/141/remotes/nextstrain.org/

The `nextstrain remote --help` output directs to that page.

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,697 @@
...
+            SubResource("application/vnd.nextstrain.dataset.main+json", ".json", primary = True),
+
+            # XXX TODO: The "sidecars" field in the /charon/getAvailable API
+            # response doesn't actually exist yet and its use here is
+            # prospective.
+            #
+            # I plan to extend the /charon/getAvailable API endpoint (or maybe
+            # switch to a new endpoint) in the future to include the "sidecars"
+            # field listing the available sidecars for each dataset, so that
+            # this code only has to try to fetch what is reported to exist.
```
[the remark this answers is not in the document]
tsibley: Yep. I imagined `ls` output to maybe initially surface this like:

```console
$ nextstrain remote ls …
…/beta-cov
…/ncov/19B (tip-frequencies, root-sequence)
…/ncov/19B/2021-02-11 (tip-frequencies, root-sequence)
…/zika (tip-frequencies)
```

similar to how `download` does.

But there's also the possibility of jumping right to something fancier, e.g. with [rich](https://github.com/Textualize/rich), as well as doing something machine-readable, e.g. JSON or YAML output mode.

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,656 @@
...
+    A remote Nextstrain dataset, as described by a Charon API response,
+    extended for the nextstrain.org RESTful API.
+    """
+    def __init__(self, api_item):
+        super().__init__(api_item)
+
+        default_sidecars = ["root-sequence", "tip-frequencies"]
+
+        self.subresources = [
+            SubResource("application/vnd.nextstrain.dataset.main+json", ".json", primary = True),
```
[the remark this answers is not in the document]
tsibley: Maybe? My thinking here was that media types are the kind of constants where the value _is_ the name and vice versa, like, say, HTTP methods. That is, they're different from constants where the value is not synonymous with the name.

For this type of self-describing constant, I think they're more clearly used directly rather than indirectly behind a translation layer (dict/enum) that gives them some other name. My thinking goes, why have multiple names, varying across code bases, for a thing when you can have one that's consistent across codebases?

One benefit of indirection with an enum (but not dict) is that it would let the type checker tell us if we made a typo, but that feels like a marginal gain not worth the indirection to me. (The code will also likely fail to work with a typo in the media type, so would be noticeable.)

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,656 @@
...
+    datasets:   Dict[str, Dict[str, Path]] = defaultdict(dict)
+    narratives: Dict[str, Dict[str, Path]] = defaultdict(dict)
+    unknowns: List[Path] = []
+
+    for path in paths:
+        # XXX TODO: In the future, in addition or instead of using filename
+        # conventions, we might want to identify files using partial content
+        # inspection (e.g. does the file contain the expected top-level keys
+        # for a given sidecar type?) or full schema validation (e.g. try to
+        # validate all schemas and see which passes).
```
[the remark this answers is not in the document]
tsibley: Agreed! Part of why I gave our existing schemas proper, resolvable ids (e.g. https://nextstrain.org/schemas/dataset/v2) is so we could leverage those here eventually. :-)

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,656 @@
...
+        status = err.response.status_code
+
+        if status in {401, 403}:
+            user = current_user()
+
+            if user:
+                raise UserError(f"""
+                    Permission denied.
+
+                    Are you logged in as the correct user?  Current user: {user.username}.
```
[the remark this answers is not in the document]
tsibley: Great idea, added in 4d2a0b9.

nextstrain/cli/remote/nextstrain_dot_org.py
tsibley: Well, "very unexpected" is not what we want! :joy: It's a good point that it's maybe a bit weird that downloading the same datasets two different ways produces different local filenames. I wonder if that situation is likely to crop up in typical usage as opposed to testing? I'd liked keeping filenames shorter than longer (e.g. instead of using the whole URL path) and picking a place to shorten them at that was under the user's control rather than picked/hardcoded by us. But maybe it's too confusing.

There's lots of brainstorming in my notes about possible behaviours here, and there were one or two decent discussions in Slack a while back, but nothing very conclusive. Maybe I should go back and summarize it all in writing and bring it back for wider consideration as part of (or separate from) this PR? Definitely would appreciate more input here.

tsibley:
> 1. It would be nice to have a `--dry-run` option for these commands.

Agreed. I'll make a new issue for that!

> 2. Should user credentials be renewed before running each command?
>    Currently, if I remove myself from a Cognito group, I can still run commands for the group as long as I don't renew my local tokens.

Maybe? This staleness is expected and part of Cognito's authn model because it issues JWTs. We can tune the maximum duration of the staleness, however. The access and id tokens are currently valid for an hour after creation and the refresh token for 30 days, which means there's up to an hour of staleness baked into an id token. We can tune it as low as 5 minutes, although this would mean very frequent background renewals.

The server side will still accept a valid token even if the CLI forces renewal before every command, so forcing a renewal keeps honest users honest ("oh, I guess I can't do that anymore") but doesn't prevent savvy/bad users from manually using the pre-renewal token ("oh, you say I can't do that anymore? I'll just use this other token that's still valid and says I can!").

We actually have a [longer staleness issue for web sessions](https://github.com/nextstrain/nextstrain.org/issues/81) right now because we don't check Cognito for updates once a session is established and we don't force periodic reauth. We should reconsider some choices there and fix that as well.

I think the more comprehensive way to deal with this staleness is to track on the server side when things like group membership changes happen and record a timestamp that says "don't accept tokens created before X for user Y", e.g. implement a basic authn revocation system.

Other than tweaking the token lifetimes, which is an ~easy knob, I think we can make the other improvements as part of further work to manage group memberships/roles within the nextstrain.org UI and API.

tsibley:
https://github.com/nextstrain/cli/issues/149

nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -0,0 +1,697 @@
...
+        # XXX TODO: Handle group-logo.png and group-overview.md by making
+        # appropriate API requests to the group endpoint (which doesn't yet
+        # exist).
+        #   -trs, 23 Sept 2021
+        raise UserError(f"""
+            Only datasets (v2) and narratives are currently supported for
+            upload to nextstrain.org, but other files were given:
+
+            {{files}}
+            """, files = file_list(unknowns))
```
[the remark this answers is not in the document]
joverlee521: Uploading unexpected files technically wouldn't "break" anything, which is why I thought this is overly cautious. 

I do agree it's not nice to leave a mess for the user to clean up. This makes me think it's insufficient that we only catch this problem sometimes. Maybe we should check this systematically and not just rely on the coincidence that they included an unknown file. 

It would be nice to get confirmation from the user when uploading files. This can be a list of the datasets and narratives or even just a summary of "Please confirm X datasets and X narratives will be uploaded". We can always add a `-f` option to allow users to bypass this confirmation step.

nextstrain/cli/remote/nextstrain_dot_org.py
joverlee521: (Ah, I just realized that I completely missed the flu example in the [download help docs](https://github.com/nextstrain/cli/pull/141/commits/a197ac4c87e63c6505e8d5785188234427f1f7aa#diff-c5bff80b5b1c30b6824da28ca7b0d22d0f9d41c8adb2c8b4b94833d664ed0bbf))

I actually prefer the opposite: keep the whole URL path so that the user can see exactly what they downloaded without having to inspect the file. They can easily rename the files locally. 

Another problem is overwriting  since the path prefixes are used to scope/filter downloads. 
If I want to download H3N2 and H1N1pdm builds to the same directory and I run
1. download nextstrain.org/flu/seasonal/h3n2/ 
2. download nextstrain.org/flu/seasonal/h1n1pdm/

The second command would just overwrite my first download 😞

joverlee521: (Relevant [Slack discussion](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1628543130011300) from a while back)

nextstrain/cli/remote/nextstrain_dot_org.py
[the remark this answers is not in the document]
tsibley: It wouldn't "break" things in the sense of "this didn't work", but it breaks things in the sense that there'd be garbage datasets/narratives listed on the Groups page, private info might be leaked, etc. And I think it breaks an expectation that these commands won't let you upload just any old files.

I agree the detection can be more systematic, hence my comments about using our schemas to do detection pre-upload instead of naive filename matching.

Confirmation is a good suggestion; I've updated #149 to describe a `--prompt` (and `--no-prompt`) mode because they're both related to `--dry-run`.

nextstrain/cli/remote/nextstrain_dot_org.py
tsibley: > I actually prefer the opposite: keep the whole URL path so that the user can see exactly what they downloaded without having to inspect the file.

I see the appeal of keeping the whole URL (though think doing `/` → `_` transformation on the whole URL is weird), but I guess I don't get why someone wouldn't know what they just downloaded.

> They can easily rename the files locally.

For a few files, yes. I'm not so confident in this being a reasonable expectation if someone does `nextstrain remote download -r nextstrain.org/ncov` (or worse, for flu) and ends up with 42 files (2 databases (GISAID, Open) × 7 regions × 3 files (main, root-sequence, tip-frequencies)).

One of the reasons for `download` using the shortest local filenames that are still distinguishing is that it helps play nice with `upload` when handling multiple datasets. Although `upload` doesn't care about local filename if uploading a single dataset, the filename still matters for multiple uploads and becomes a suffix on the given destination URL.

I'm imagining that a download + upload combo would be used for copying files between Groups, but maybe that would be less common than "download, do secondary analyses, never upload back", for which you'd tend to want longer more descriptive filenames. (The primary usage is still "run a build, upload files", and `download` doesn't matter for that.)

tsibley: > Another problem is overwriting since the path prefixes are used to scope/filter downloads. If I want to download H3N2 and H1N1pdm builds to the same directory and I run
> 
>     1. download nextstrain.org/flu/seasonal/h3n2/
> 
>     2. download nextstrain.org/flu/seasonal/h1n1pdm/
> 
> 
> The second command would just overwrite my first download

Yeah, this is not ideal. The workaround is downloading each into a specific directory (or, in the nearish future, giving a filename prefix instead of a dir).

This comes back to the Slack discussion from the fall, where you pointed out that providing a toggle between two different ways of naming local files may be best. What is better for the "download then re-upload" workflow is not better for the "download and analyze" workflow and vice versa.

PR #145, opened 2022-02-04
runner.aws_batch: Exit 1 if the job failed but there's no job exit code

tsibley:
Failed jobs do not have an exit code in Batch if the job failed for
infrastructural or reasons internal to AWS rather than job-specific
reasons.  For example, a job that fails because its EC2 instance is
shutdown in the middle of it will not have an exit code.

While `nextstrain build` used to exit 0 with success due to the missing
exit code, even though it correctly reported the job's failure and
failure reason.  Now it will exit 1 for a generic failure.

### Testing
Tested by attaching to 3 different kinds of already done jobs we had lying around:

1. Failed job where the build workflow exited 1. `nextstrain build` exits 1, as usual.
2. Failed job where the EC2 instance it was running on was terminated unexpectedly. `nextstrain build` now exits 1 with this PR.
3. Successful job where the build workflow exited 0. `nextstrain build` exits 0, as usual.

### DIFF
--- a/nextstrain/cli/runner/aws_batch/__init__.py
+++ b/nextstrain/cli/runner/aws_batch/__init__.py
@@ -267,8 +267,13 @@ def run(opts, argv, working_volume = None, extra_env = {}, cpus: int = None, mem
         s3.download_workdir(remote_workdir, local_workdir, patterns)
 
 
-    # Exit with the job's exit code, or assume success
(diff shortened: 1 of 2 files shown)


tsibley:
CI failures appear completely unrelated. Will investigate those when I have time.

victorlin (approved):

PR #147, opened 2022-02-04
Fix external test breakage

tsibley:
See commit messages. If CI passes for this PR, then the PR has done its job. :-)


tsibley:
Ok, now just 3.9 is failing with:

    ModuleNotFoundError: No module named 'importlib.readers'

from inside pytest/importlib, which is… odd. Will try to repro locally and see what's up.

PR #157, opened 2022-03-03
config: Use a global lockfile separate from the config/secrets files

### LINKED ISSUE #151: `[Errno 13] Permission denied` on Windows
_from [discussion post](https://discussion.nextstrain.org/t/nextstrain-cli-installation-permission-error-errno-13/1011)_

**Current Behavior**  

The error happens with at least the following 3 commands:

1. `nextstrain update`

    ```
    >nextstrain update
    ...
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\Scripts\nextstrain-script.py", line 9, in <module>
        sys.exit(main())
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__main__.py", line 10, in main
        return cli.run( argv[1:] )
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__init__.py", line 31, in run
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\command\update.py", line 25, in run
        statuses = [
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\command\update.py", line 26, in <listcomp>
        runner.update()
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\runner\docker.py", line 234, in update
        config.set("docker", "image", latest_image)
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 105, in set
        save(config, path)
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 72, in save
        config.write(file)
    PermissionError: [Errno 13] Permission denied
    ```

2. `nextstrain check-setup --set-default` with docker, aws-batch supported

    ```
    >nextstrain check-setup --set-default
    ...
    Setting default environment to docker.
    PermissionError: [Errno 13] Permission denied

    During handling of the above exception, another exception occurred:

    File "C:\Users\vlin\Miniconda3\envs\nextstrain\Scripts\nextstrain-script.py", line 9, in <module>
        sys.exit(main())
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__main__.py", line 10, in main
        return cli.run( argv[1:] )
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__init__.py", line 31, in run
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\command\check_setup.py", line 108, in run
        config.set("core", "runner", supported_runners[0])
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 105, in set
        save(config, path)
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 72, in save
        config.write(file)
    PermissionError: [Errno 13] Permission denied
    ```

3. `nextstrain check-setup --set-default` with aws-batch supported

    ```
    >nextstrain check-setup --set-default
    ...
    All good!  Supported Nextstrain environments: aws-batch

    Setting default environment to aws-batch.
    PermissionError: [Errno 13] Permission denied

    During handling of the above exception, another exception occurred:

    File "C:\Users\vlin\Miniconda3\envs\nextstrain\Scripts\nextstrain-script.py", line 9, in <module>
        sys.exit(main())
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__main__.py", line 10, in main
        return cli.run( argv[1:] )
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\__init__.py", line 31, in run
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\command\check_setup.py", line 108, in run
        config.set("core", "runner", supported_runners[0])
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 105, in set
        save(config, path)
    File "C:\Users\vlin\Miniconda3\envs\nextstrain\lib\site-packages\nextstrain\cli\config.py", line 72, in save
        config.write(file)
    PermissionError: [Errno 13] Permission denied
    ```

**Expected behavior**  

Normal behavior.

**How to reproduce**  
Steps to reproduce the current behavior:  
1. Install Nextstrain using the instructions for Windows OS in [install docs](https://docs.nextstrain.org/en/latest/install.html).
2. Run any one of the commands above.
3. See error.

**Possible solution**  
(optional)  

**Your environment: if running Nextstrain locally**  
 - Operating system: Windows 10
 - Version: nextstrain-cli 3.0.6

**Additional context**  

I checked on 3.0.3, 3.0.4, 3.0.5, and 3.0.6. The last working version was 3.0.3. So the bug was introduced somewhere here: https://github.com/nextstrain/cli/compare/3.0.3...3.0.4

Workaround: use WSL on Windows to install Nextstrain.

tsibley:
Locking the same file we then re-open for writing doesn't work on
Windows because the underlying LockFileEx is fd/handle-scoped not
process-scoped and a mandatory, OS enforced-lock, as documented¹:

> If the locking process opens the file a second time, it cannot access
> the specified region through this second handle until it unlocks the
> region.

The pattern works on Unix because the fcntl()-based lock isn't
mandatory.  This difference is noted in fastener's README², but I didn't
realize the full implications of Windows' mandatory lock when
introducing the locking in 821c08e.  My thinking was enforcement was
process-scoped not fd/handle-scoped.

Using a global lockfile instead of a per-file sidecar lockfile avoids
having to reason about the parent path existence checks which is a) hard
and b) creates several small race conditions (which were present until
now).  The downside of this simpler implementation is that accesses to
different files will be bottlenecked on the same lock (e.g. a writer to
secrets will block a reader of config; a reader of config will block a
writer of secrets).  This is just fine, however, given our usage and
very short holding times of the locks.

Resolves #151, which also discusses a rejected alternative of re-using
the same fd that fasteners locks.

¹ https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex
² https://pypi.org/project/fasteners/0.17.3/#description

### DIFF
--- a/nextstrain/cli/config.py
+++ b/nextstrain/cli/config.py
@@ -26,6 +26,10 @@ CONFIG = Path(os.environ.get("NEXTSTRAIN_CONFIG") or
 SECRETS = Path(os.environ.get("NEXTSTRAIN_SECRETS") or
                HOME / ".nextstrain/secrets")
 
+# Path to our global lock file
+LOCK = Path(os.environ.get("NEXTSTRAIN_LOCK") or
+            HOME / ".nextstrain/lock")
+
 # Permissions to use for the secrets file if we have to create it.
 SECRETS_PERMS = \
     ( stat.S_IRUSR  # u+r
@@ -56,7 +60,6 @@ def save(config, path: Path = CONFIG):
     """
     secrets = path is SECRETS
 
-    # See also the handling of parents in write_lock().
     path = path.resolve(strict = False)
 
     if path.parent.name == ".nextstrain":
@@ -79,7 +82,7 @@ def get(section: str, field: str, fallback: str = None, path: Path = CONFIG) -> 
     If *section* or *field* does not exist, returns *fallback* (which defaults
     to None).
     """
-    with read_lock(path):
+    with read_lock():
         config = load(path)
 
         if section in config:
@@ -94,7 +97,7 @@ def set(section: str, field: str, value: str, path: Path = CONFIG):
 
     If *section* does not exist, it is automatically created.
     """
-    with write_lock(path):
+    with write_lock():
         config = load(path)
 
         if section not in config:
@@ -115,7 +118,7 @@ def remove(section: str, path: Path) -> bool:
     if not path.exists():
         return False
 
-    with write_lock(path):
+    with write_lock():
         config = load(path)
 
         if section not in config:
@@ -129,38 +132,22 @@ def remove(section: str, path: Path) -> bool:
 
 
 @contextmanager
-def read_lock(path: Path):
+def read_lock():
     """
-    Lock *path* for reading across processes (but not within).
+    Lock for reading across processes (but not within).
 
-    Uses advisory/cooperative locks.  Avoids creating *path* if it does not
-    exist.
+    Uses a global advisory/cooperative lock.
     """
-    if path.exists():
-        with InterProcessReaderWriterLock(path).read_lock():
-            yield
-    else:
+    with InterProcessReaderWriterLock(LOCK).read_lock():
         yield
 
 
 @contextmanager
-def write_lock(path: Path):
+def write_lock():
     """
-    Lock *path* for writing across processes (but not within).
+    Lock for writing across processes (but not within).
 
-    Uses advisory/cooperative locks.  Avoids creating *path* if the parent does
-    not exist.
+    Uses a global advisory/cooperative lock.
     """
-    # We only care if the parent exists, as the file is expected to be written
-    # and created if it does not exist.
-    parent = path.resolve(strict = False).parent
-
-    if parent.name == ".nextstrain":
-        # save() will auto-create in this case, so look one parent higher.
-        parent = parent.parent
-
-    if parent.exists():
-        with InterProcessReaderWriterLock(path).write_lock():
-            yield
-    else:
+    with InterProcessReaderWriterLock(LOCK).write_lock():
         yield


victorlin (approved):
Nice job tracking this down without access to a Windows machine! Tested on Windows 10 with https://github.com/nextstrain/cli/issues/151#issuecomment-1058502142 and this solves the issue.

>  The downside of this simpler implementation is that accesses to
different files will be bottlenecked on the same lock

I'm not sure where these locks take place and for how long, but would it result in scenarios where multiple SLURM jobs are submitted under the same user, happen to run `nextstrain update` at the same time, causing all but one job to crash?

If the chances of that happening are slim, then this is perfect. If not, then some sort of error handling or retry would be better than `[Errno 13] Permission denied`.

tsibley:
> would it result in scenarios where multiple SLURM jobs are submitted under the same user, happen to run `nextstrain update` at the same time, causing all but one job to crash?
> 
> If the chances of that happening are slim, then this is perfect. If not, then some sort of error handling or retry would be better than `[Errno 13] Permission denied`.

Ah, so the effect of the locking is writes are serialized and reads wait until a write is finished, at which point reads can proceed in parallel. Lock acquisition retries automatically, as the default acquisition method is blocking without a timeout.

The errno 13 crash shouldn't happen again; it was the result of the mandatory lock denying access to an operation that didn't hold the lock, not the result of lock acquisition failing or waiting.

Re: SLURM specifically, if the jobs are accessing a shared filesystem that supports `fcntl()` locking (like modern NFS) then it'll all work out fine. If the shared filesystem doesn't support locking (:thinking:), either it'll silently not actually lock or throw an "unsupported" error for _all_ lock acquisitions and none will succeed. IME the former is more common (:scream:). I'm not worried about this case though; it should be very rare to occur, if at all, because of the timing requirements and on top of that, modern shared FS support locks.

Did you have a specific use case in mind with the SLURM question, or was it just a general example?

victorlin:
@tsibley

> Did you have a specific use case in mind with the SLURM question, or was it just a general example?

No, was just trying to think of ways where a global lock could be problematic. But like you've described, we should be good here!

PR #153, opened 2022-02-26
[check-setup] Improve memory limit check

tsibley:
### Description of proposed changes    

- Improves the error handling of the check
- Supports host systems with cgroups v2
- Uses host limit even if cgroups limit isn't available

See commit messages for details.

### Related issue(s)  
Fixes #135.

### Testing
I tested using the [local VM setup I described](https://github.com/nextstrain/cli/issues/135#issuecomment-1051386199) in #135 by first reproducing the issues with nextstrain-cli 3.0.6 from PyPI and then installing this branch to see the issues resolved:

```console
$ pip3 install --user 'nextstrain-cli @ https://github.com/nextstrain/cli/archive/trs/cgroups-v2.zip'
…
$ ~/.local/bin/nextstrain check-setup
```

I also tested each commit individually to make sure they each worked as intended.

PR #159, opened 2022-03-05
Handle memory limits reported from Docker as floats

huddlej:
## Context

The fix in #153 produced a new error for me with Docker 4.5.0 on OS X 10.14.6:

```bash
$ python3 -m pip install --upgrade nextstrain-cli
$ nextstrain --version
nextstrain.cli 3.1.1
$ cat ~/.nextstrain/config
[docker]
image = nextstrain/base:build-20220215T000459Z

[core]
runner = native
$ nextstrain check-setup --set-default
nextstrain-cli is up to date!

Testing your setup…
  File "/Users/jlhudd/miniconda3/envs/nextstrain/bin/nextstrain", line 8, in <module>
    sys.exit(main())
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/__main__.py", line 10, in main
    return cli.run( argv[1:] )
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/__init__.py", line 33, in run
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/command/check_setup.py", line 63, in run
    runner_tests = [
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/command/check_setup.py", line 64, in <listcomp>
    (runner, runner.test_setup())
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/runner/docker.py", line 226, in test_setup
    *test_memory_limit(),
  File "/Users/jlhudd/miniconda3/envs/nextstrain/lib/python3.8/site-packages/nextstrain/cli/runner/docker.py", line 172, in test_memory_limit
    limit = min(filter(None, map(int_or_none, limits)))
ValueError: min() arg is an empty sequence
```

Inspecting the values of `limits` in pdb, I find:

```python
ipdb> limits
['4.1257e+09', 'max']
ipdb> print(list(map(int_or_none, limits)))
[None, None]
```

Since the OS reports the memory limit with an exponential value, `int` cannot interpret the string and returns a `ValueError`. However, `float` can interpret this string, so `int(float("4.1257e+09"))` produces `4125700000`, as expected.

After changing `int_or_none` accordingly, I ran the `nextstrain` command again (this time from a local dev installation with my changes).

```
$ nextstrain check-setup --set-default
$ cat ~/.nextstrain/config
[docker]
image = nextstrain/base:build-20220215T000459Z

[core]
runner = docker
```

And the command ran as expected.

Fixes the issue described above on Mac OS X with Docker where `check-setup --set-default` fails when the memory limits reported from the Docker OS are in floating point values (e.g., "") that cannot be cast to integers by Python. This commit addresses the issue by first casting the string to float before casting to int.

### Related issue(s)

Related to #153

### Testing

 - [x] Tested locally to confirm the issue was fixed for me
 - [ ] Tested by CI

### DIFF
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -165,7 +165,7 @@ def test_setup() -> RunnerTestResults:
             else:
                 def int_or_none(x):
                     try:
-                        return int(x)
+                        return int(float(x))
                     except ValueError:
                         return None
 
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -184,6 +184,7 @@ def test_setup() -> RunnerTestResults:
                             """ % (limit / GiB))
                         status = None
                     else:
+                        msg += " (limit is %.1f GiB)" % (limit / GiB)
                         status = True
 
         return [(msg, status)]
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -169,19 +169,22 @@ def test_setup() -> RunnerTestResults:
                     except ValueError:
                         return None
 
-                limit = min(filter(None, map(int_or_none, limits)))
+                limits = list(filter(None, map(int_or_none, limits)))
 
-                if limit <= desired:
-                    msg += dedent("""
+                if limits:
+                    limit = min(limits)
 
-                        Containers appear to be limited to %0.1f GiB of memory. This
-                        may not be enough for some Nextstrain builds.  On Windows or
-                        a Mac, you can increase the memory available to containers
-                        in the Docker preferences.\
-                        """ % (limit / GiB))
-                    status = None
-                else:
-                    status = True
+                    if limit <= desired:
+                        msg += dedent("""
+
+                            Containers appear to be limited to %0.1f GiB of memory. This
+                            may not be enough for some Nextstrain builds.  On Windows or
+                            a Mac, you can increase the memory available to containers
+                            in the Docker preferences.\
+                            """ % (limit / GiB))
+                        status = None
+                    else:
+                        status = True
 
         return [(msg, status)]
 
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -153,24 +153,22 @@ def test_setup() -> RunnerTestResults:
         status: RunnerTestResultStatus = ...
 
         if image_exists():
+            def int_or_none(x):
+                try:
+                    return int(float(x))
+                except ValueError:
+                    return None
+
             report_memory = """
                 awk '/^MemTotal:/ { print $2 * 1024 }' /proc/meminfo
                 (cat /sys/fs/cgroup/memory.max || cat /sys/fs/cgroup/memory/memory.limit_in_bytes) 2>/dev/null
             """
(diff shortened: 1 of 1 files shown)


tsibley:
Ugh, sorry about that. Thanks for this fix. I think we should also address the underlying `ValueError: min() arg is an empty sequence` issue so it doesn't remain latent until a new reason causes it. Will push another commit to this branch.

tsibley:
I was incredulous that /proc/meminfo reported this in exponential notation, so I dug a bit deeper. Indeed, [`proc(7)`]() documents `MemTotal` to be a long unsigned int (`%lu`), like most fields in this file. The culprit is awk (!):

```
nextstrain:/nextstrain/build $ awk '/^MemTotal:/ { print $2 * 1024 }' /proc/meminfo
1.657e+10
nextstrain:/nextstrain/build $ awk '/^MemTotal:/ { print $2 }' /proc/meminfo
16181632
```

Locally this doesn't happen, so presumably there's some newer/different version of awk in the container than I have or maybe some locale-dependent thing. But this [as-documented](https://www.gnu.org/software/gawk/manual/gawk.html#Controlling-Numeric-Output-with-print):

> The predefined variable `OFMT` contains the format specification that print uses with `sprintf()` when it wants to convert a number to a string for printing. The default value of `OFMT` is "`%.6g`".

The upshot is we could fix this by instead doing the right thing in awk:

```
printf "%lu\n", $2 * 1024
```

But `int(float(…))` in Python is also fine.

tsibley:
@huddlej Pushed two additional commits addressing my other concern. Would you take a look?

huddlej:
> Maybe worth ditching the vagaries of awk's numeric handling, even if the exponential notation it produces seems ok on the face of it?

Agreed about avoiding awk vagaries where possible. It's been ages since I used Perl; will it avoid creating exponential values for the conversion to bytes? The other solution that seems fine is to make two separate system calls to parse `/proc/meminfo` and the cgroups info and handle the math in Python.

> Pushed two additional commits addressing my other concern. Would you take a look?

These changes look good to me! Also agree it's better to protect against the empty list in the future.

tsibley:
> Agreed about avoiding awk vagaries where possible. It's been ages since I used Perl; will it avoid creating exponential values for the conversion to bytes? The other solution that seems fine is to make two separate system calls to parse `/proc/meminfo` and the cgroups info and handle the math in Python.

I'm going to leave it as-is with awk for now to avoid accidentally breaking it another way. I didn't mean to suggest replacing awk with Perl here. Perl will generally avoid creating exponential notation… but I believe only up to a certain point (that's larger than awk's).

Two separate `run_bash()` calls aren't enticing because they're a bit slow, but yeah, that's an option. We could also run Python in the container instead and have it spit out the final limit. But I'll save one of those for another day and merge this now (once CI passes) to fix the immediate issue.

tsibley:
@huddlej Released with 3.2.0, so you can install that and make `check-setup` work again!

huddlej:
@tsibley It works! Thank you!

Issue #14, opened 2018-07-24
Automated testing on Windows


tsibley:
This is possible with AppVeyor, but not using their free tier because we need to run Linux containers on Windows. Nevertheless, the proof-of-concept is: https://github.com/nextstrain/cli/compare/windows-testing

victorlin:
This is tricky because we need to install `augur` and `auspice` for the zika-tutorial test, but there is no way to properly install those on Windows:

1. `augur` has runtime dependencies (e.g. [mafft](https://anaconda.org/bioconda/mafft), [iqtree](https://anaconda.org/bioconda/iqtree)) that are not available on Windows, at least via conda
2. `auspice` is [not availabe on Windows via conda](https://anaconda.org/bioconda/auspice) (can this be changed?)

See [example run](https://github.com/nextstrain/cli/runs/5485533209?check_suite_focus=true):

```
Encountered problems while solving:
  - nothing provides requested auspice
  - nothing provides mafft needed by augur-10.0.0-py_0
```

tsibley:
We may be able to package Auspice for Windows. However, Augur still remains an issue. We can skip building zika-tutorial with `--native` on Windows, though? This would still test one of our recommended install method on Windows and cover a lot of ground. Relatedly, it'd be good to test the other Windows install method too inside WSL.

tsibley:
Notes from some quick looking again at CI for Windows.

## Docker runtime

Neither the `windows-2022` or `windows-2019` runners that GitHub Actions provides support WSL 2 or Hyper-V, so while Docker is installed and "works", it appears limited to running Windows containers. Bummer!

We could potentially address this by self-hosting a runner on a Windows machine, or looking at other CI services which might provide this (e.g. my unfinished AppVeyor proof-of-concept from 2018).

## Native runtime

Bioconda by policy [doesn't build packages for Windows](https://bioconda.github.io/user/versions.html#operating-systems), even if it would be possible for a package.

Conda Forge does, however, so one option is to submit/maintain recipes there for a few of our external deps to make it easier to bootstrap a native runtime on Windows.

Otherwise, here's what currently exists.

### MAFFT

[...]

PR #160, opened 2022-03-09
Update CI workflow

victorlin:

This PR includes some changes to the GitHub Actions workflow used for CI, without any functional difference. See individual commit messages.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,9 @@ jobs:
           - 3.8
           - 3.9
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash -l -eo pipefail {0}
     steps:
       - uses: actions/checkout@v2
       - uses: conda-incubator/setup-miniconda@v2
@@ -28,6 +31,8 @@ jobs:
           channels: conda-forge,bioconda
           channel-priority: true
           activate-environment: test
+      - run: cat ~/.profile || true
+      - run: cat ~/.bashrc || true
 
       - name: Set up test environment
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,43 +21,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: ${{ matrix.python }}
+          mamba-version: "*"
+          channels: conda-forge,bioconda
+          channel-priority: true
+          activate-environment: test
 
-      - name: Upgrade Python toolchain
-        run: python3 -m pip install --upgrade pip setuptools wheel
-
-      - name: Setup test environment
+      - name: Set up test environment
         run: |
-          # Install required external programs for --native.
-          conda create -p ~/ci -c conda-forge -c bioconda \
-              fasttree \
-              iqtree \
-              mafft \
-              nodejs=12 \
-              raxml \
-              vcftools
-
-          # Add Conda env to PATH for this and subsequent steps
-          export PATH="$HOME/ci/bin:$PATH"
-          tee -a "$GITHUB_PATH" <<<"$HOME/ci/bin"
-
-          # Install required Python programs for --native, intentionally
-          # outside of the Conda env since otherwise they'll drag Python into
-          # the Conda env and shadow our desired Python version from the matrix
-          # and setup-python above.
-          python3 -m pip install \
-              nextstrain-augur \
-              snakemake
-
-          # Installs into Conda env to avoid version/permissions issues.
-          npm install --global auspice
-
+          mamba install augur auspice snakemake
           # The small build we'll use as an integration test case.
           git clone https://github.com/nextstrain/zika-tutorial
-        env:
-          CONDA_PKGS_DIRS: /tmp
 
       - name: Install Nextstrain CLI
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,13 +35,17 @@ jobs:
           # The small build we'll use as an integration test case.
           git clone https://github.com/nextstrain/zika-tutorial
 
+      - name: Install Nextstrain CLI
+        run: python3 -m pip install --upgrade '.[dev]'
+
+      - run: conda info
+      - run: conda list
+
       - name: Check python version
         run: |
           # Assert that we're on the expected Python version, in case the GH
           # Actions environment is messed up.
           python3 --version | grep -F 'Python ${{ matrix.python }}.'
-      - name: Install Nextstrain CLI
-        run: python3 -m pip install --upgrade '.[dev]'
 
       - name: Run Nextstrain CLI commands
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,10 +14,10 @@ jobs:
           - ubuntu-latest
           - macos-latest
         python:
-          - 3.6
-          - 3.7
-          - 3.8
-          - 3.9
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
     runs-on: ${{ matrix.os }}
     defaults:
       run:
@@ -84,7 +84,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: '3.6'
 
       - name: Upgrade Python toolchain
         run: python3 -m pip install --upgrade pip setuptools wheel
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,13 +35,16 @@ jobs:
           # The small build we'll use as an integration test case.
           git clone https://github.com/nextstrain/zika-tutorial
 
-      - name: Install Nextstrain CLI
+      - name: Check python version
         run: |
           # Assert that we're on the expected Python version, in case the GH
           # Actions environment is messed up.
           python3 --version | grep -F 'Python ${{ matrix.python }}.'
-          python3 -m pip install --upgrade '.[dev]'
+      - name: Install Nextstrain CLI
+        run: python3 -m pip install --upgrade '.[dev]'
 
+      - name: Run Nextstrain CLI commands
+        run: |
           nextstrain version --verbose
           nextstrain check-setup
 


.github/workflows/ci.yaml
```diff
@@ -35,14 +35,12 @@ jobs:
           # The small build we'll use as an integration test case.
           git clone https://github.com/nextstrain/zika-tutorial
 
-      - name: Check python version
-        run: |
-          # Assert that we're on the expected Python version, in case the GH
-          # Actions environment is messed up.
-          python --version | grep -F 'Python ${{ matrix.python }}.'
```
tsibley: (from the commit message)

> Grepping for python version won't actually raise an error when not found.
> This means it is only useful for debugging when looking at run logs.

Ah, no, that's not right. `grep` will exit 1 when it doesn't find a match, which will cause the job step to fail because [the default `shell` setting](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) includes bash's `-eo pipefail` options.

.github/workflows/ci.yaml
```diff
@@ -19,6 +19,9 @@ jobs:
           - 3.8
           - 3.9
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash -l {0}
```
tsibley, line 24: [The default](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) is

```
bash --noprofile --norc -eo pipefail {0}
```

so this change effectively removes the `-eo pipefail` options. We want to keep them.

Hopefully by switching to a login shell (`-l`) and thus sourcing the profile and rc files we don't bring in other unexpected changes besides whatever Conda added. This is why I just explicitly set up `PATH` instead.

.github/workflows/ci.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: Run pytest
         run: pytest -v
 
-      - if: runner.os != 'macOS'
+      - if: matrix.os != 'macos-latest'
```
tsibley, line 61: > matrix.os is easier to understand as it is a direct comparison with
> the values provided in the matrix definition.

Not a bit deal, but the downside is that this will now have to be updated if/when we add other macOS versions to the matrix. I bet we'll forget… ;-) My use of `runner.os` here was intentional for this reason.

.github/workflows/ci.yaml
```diff
@@ -14,10 +14,10 @@ jobs:
...
           - macos-latest
         python:
-          - 3.6
-          - 3.7
-          - 3.8
-          - 3.9
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
```
tsibley, line 20: :+1:

tsibley (changes requested):
Thanks for working to improve this CI! I'm supportive of making the setup/scaffolding better/simpler/etc but have a few questions and a change or two to address.

tsibley:
Two additional notes:

1. I have some slight concern that by running inside a Conda env into which we also install deps that rely on Python, one of those deps will pull in a newer Python version than `matrix.python` and we won't notice it. This would mean we'd quietly stop testing the Python version we mean to.

2. What do we make of this warning that all jobs now emit?
   > Mamba support is still experimental and can result in differently solved environments!

victorlin:
@tsibley thanks for reviewing! I started on this journey to address #14 but ran into other issues which I'll write up in a comment there. Motivation was mainly when I saw the bash-specific usage (`export PATH`, `<<<`), I thought using conda/mamba would have more cross-platform potential and be consistent with [augur's CI workflow](https://github.com/nextstrain/augur/blob/master/.github/workflows/ci.yaml).

> I have some slight concern that by running inside a Conda env into which we also install deps that rely on Python, one of those deps will pull in a newer Python version than matrix.python and we won't notice it.

Mamba automatically pins Python to the minor version, as seen in output:

```
Pinned packages:
  - python 3.6.*
```

I couldn't find any good documentation on this behavior but it is implemented [here](https://github.com/mamba-org/mamba/blob/bff16c2bdc4103ba74c23ab4fdbf58849a55981c/libmamba/src/core/pinning.cpp#L14).

> What do we make of this warning that all jobs now emit?
> > Mamba support is still experimental and can result in differently solved environments!

In my (limited) experience mamba has been fine, and I haven't heard of any issues from other teammates who also use it. We recommend it in the [install docs](https://docs.nextstrain.org/en/latest/install.html) and is already used in [augur's CI workflow](https://github.com/nextstrain/augur/blob/master/.github/workflows/ci.yaml) which has been fine while emitting these warnings. But yes, it is still in `0.x.x` development stage. Some other options here:

1. Use conda. Downside: slow
2. Specify package versions such as `augur=14 auspice=2 snakemake=6`. Downside: we would have to bump these manually, and it wouldn't catch potential breakage with latest versions.

.github/workflows/ci.yaml
victorlin: I see `'macOS'` as an additional value to look up in documentation. But this change isn't necessary so I'll revert.

.github/workflows/ci.yaml
victorlin: The problem with `PATH` is it calls for a separate script for Windows if we want to support that. But we might have to do that regardless, to some extent.

I'll:

1. Add `-eo pipefail` to this default shell
2. Add commands to write the contents of `~/.profile` and `~/.bashrc` so we can make sure nothing unwanted is in there.

.github/workflows/ci.yaml
victorlin: Oh that makes sense now! I was trying to understand why [this was failing](https://github.com/nextstrain/cli/runs/5484946805?check_suite_focus=true#step:5:18) (before I used `bash -l {0}` so it was not activating the conda environment hence different version).

In that case, better to add the `grep` check back.

.github/workflows/ci.yaml
```diff
@@ -35,13 +35,16 @@ jobs:
...
         run: |
           # Assert that we're on the expected Python version, in case the GH
           # Actions environment is messed up.
           python --version | grep -F 'Python ${{ matrix.python }}.'
-          python -m pip install --upgrade '.[dev]'
+      - name: Install Nextstrain CLI
+        run: python -m pip install --upgrade '.[dev]'
 
+      - name: Run Nextstrain CLI commands
+        run: |
```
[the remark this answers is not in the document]
victorlin: I was trying to understand why [this was failing](https://github.com/nextstrain/cli/runs/5484946805?check_suite_focus=true#step:5:18). I think it's more clear [here](https://github.com/nextstrain/cli/runs/5485205949?check_suite_focus=true#step:5:8) where it fails with `python --version` being the only command, though also having the output of `python --version` would be nice too.

.github/workflows/ci.yaml
```diff
@@ -21,50 +21,26 @@ jobs:
...
-          CONDA_PKGS_DIRS: /tmp
 
       - name: Install Nextstrain CLI
         run: |
           # Assert that we're on the expected Python version, in case the GH
           # Actions environment is messed up.
-          python3 --version | grep -F 'Python ${{ matrix.python }}.'
-          python3 -m pip install --upgrade '.[dev]'
+          python --version | grep -F 'Python ${{ matrix.python }}.'
+          python -m pip install --upgrade '.[dev]'
```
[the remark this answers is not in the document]
victorlin: Stylistic change I made while figuring out something else. Not necessary so I'll revert.

.github/workflows/ci.yaml
```diff
@@ -21,50 +21,26 @@ jobs:
...
+      - uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: ${{ matrix.python }}
+          mamba-version: "*"
+          channels: conda-forge,bioconda
+          channel-priority: true
+          activate-environment: test
 
-      - name: Upgrade Python toolchain
-        run: python3 -m pip install --upgrade pip setuptools wheel
```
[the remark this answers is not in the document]
victorlin: I think so, at least if installing through `conda-forge`:

```
% conda search -c conda-forge python=3.6 --info
...
dependencies:
    ...
  - pip
```

```
% conda search -c conda-forge pip --info
...
dependencies: 
  ...
  - setuptools
  - wheel
```

The lack of pinned version numbers would resolve to latest versions. Though this has weird side effects such as https://github.com/nextstrain/augur/pull/840#issuecomment-1028220873

.github/workflows/ci.yaml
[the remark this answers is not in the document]
tsibley: Ok to leave too if you want; I don't materially care either way, was just curious!

.github/workflows/ci.yaml
tsibley: Sounds good!

.github/workflows/ci.yaml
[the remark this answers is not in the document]
victorlin: Also, having more modular steps makes it easier to find specific output on GitHub's UI ([example](https://github.com/nextstrain/cli/runs/5486106181?check_suite_focus=true))

.github/workflows/ci.yaml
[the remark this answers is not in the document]
victorlin: Reverted – better to be consistent with the `python3` already used elsewhere in this file!

.github/workflows/ci.yaml
victorlin: The [new CI run](https://github.com/nextstrain/cli/actions/runs/1960723783) shows that there is no `~/.bashrc` on these runners. Here are the contents of `~/.profile` on each os:

ubuntu-latest:

```
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
__conda_setup="$('/usr/share/miniconda/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__conda_setup"
else
    if [ -f "/usr/share/miniconda/etc/profile.d/conda.sh" ]; then
        . "/usr/share/miniconda/etc/profile.d/conda.sh"
    else
        export PATH="/usr/share/miniconda/bin:$PATH"
    fi
fi
unset __conda_setup
# <<< conda initialize <<<


  # ----------------------------------------------------------------------------
  # Conda Setup Action: Basic configuration
  set -eo pipefail
  # Conda Setup Action: Custom activation
  conda activate "test"
  # ----------------------------------------------------------------------------
```

macos-latest:

```
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
__conda_setup="$('/usr/local/miniconda/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__conda_setup"
else
    if [ -f "/usr/local/miniconda/etc/profile.d/conda.sh" ]; then
        . "/usr/local/miniconda/etc/profile.d/conda.sh"
    else
        export PATH="/usr/local/miniconda/bin:$PATH"
    fi
fi
unset __conda_setup
# <<< conda initialize <<<


  # ----------------------------------------------------------------------------
  # Conda Setup Action: Basic configuration
  set -eo pipefail
  # Conda Setup Action: Custom activation
  conda activate "test"
  # ----------------------------------------------------------------------------
```

Mostly all conda stuff which is wanted. The conda action also adds `set -eo pipefail`, but good to have it defined in our default shell in case they remove it here.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
tsibley: Good to know, thanks!

.github/workflows/ci.yaml
tsibley: Ah, nice to see them restore `-eo pipefail` since they document switching the shell to just `bash -l`.

tsibley:
Nod. I'm supportive of switching the whole test setup into a Conda/Mamba env, esp. if it makes the setup for Windows testing easier.

> Mamba automatically pins Python to the minor version, as seen in output:

Ahh, great, TIL. Thanks. It looks like this pinning is [also done separately](https://github.com/mamba-org/mamba/blob/bff16c2bdc4103ba74c23ab4fdbf58849a55981c/mamba/mamba/mamba_env.py#L60-L69) inside `mamba env {create, update}`. (`mamba::python_pin()` is looks unused by the `mamba` Python frontend; it looks used only by the higher-level libmamba C++ API used by `micromamba`.)

Best "doc" I could find for this pinning behaviour are discussions in two issues:

- https://github.com/mamba-org/mamba/issues/1245
- https://github.com/mamba-org/mamba/issues/1002

> In my (limited) experience mamba has been fine, and I haven't heard of any issues from other teammates who also use it. We recommend it in the [install docs](https://docs.nextstrain.org/en/latest/install.html) and is already used in [augur's CI workflow](https://github.com/nextstrain/augur/blob/master/.github/workflows/ci.yaml) which has been fine while emitting these warnings.

Great! That we haven't encountered issues related to the warning while using it in Augur answers my question.

tsibley (approved):

PR #161, opened 2022-03-22
remote upload: Skip gzip compression for files that are already compressed

tsibley:
This avoids some useless work but mostly serves to head off confusion
(e.g. curl without the --compression option) and/or quirks of HTTP
clients (e.g. Snakemake's HTTP remote file provider¹) when a compressed
file is compressed again and served with a Content-Encoding: gzip
header.²

This doesn't come up with Nextstrain dataset and narrative files but
does with adjacent input files like metadata.tsv.gz and
sequences.fasta.xz which we also put in the S3 buckets (e.g.
s3://nextstrain-data/files/zika/…).³

The remote family of commands is not intended for generic S3 management
per se, but they're often useful in the Nextstrain ecosystem to manage
these ancillary data files.  Part of this is that the commands are handy
and available, part of it is that Cloudfront invalidation still remains
a complication with using `aws s3` directly.  Avoiding double
compression doesn't go so far out of our way and helps support this
slightly off-label use case.

¹ https://github.com/snakemake/snakemake/issues/1508
² https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1647910842228169
³ https://github.com/nextstrain/fauna/pull/114

### DIFF
--- a/nextstrain/cli/remote/s3.py
+++ b/nextstrain/cli/remote/s3.py
@@ -52,7 +52,7 @@ from operator import methodcaller
 from os.path import commonprefix
 from pathlib import Path
 from time import time
-from typing import Iterable, List, Tuple
+from typing import Iterable, List, Optional, Tuple
 from .. import aws
 from ..gzip import GzipCompressingReader, ContentDecodingWriter
 from ..util import warn, remove_prefix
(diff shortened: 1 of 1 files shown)


trvrb (approved):

trvrb:
Thanks for including these guardrails @tsibley. Running with my original commands prevents the confusing behavior. Much appreciated.

tsibley:
Released with 3.2.1.

Issue #114, opened 2021-03-05
Detect install from Conda package in upgrade hints


tsibley:
Now that there's a [nextstrain-cli](https://bioconda.github.io/recipes/nextstrain-cli/README.html) package in Bioconda, the [hint in `check_for_new_version()`](https://github.com/nextstrain/cli/blob/42be967c936846a54f300f31842cf7804ce3e428/nextstrain/cli/util.py#L52-L84) should detect when its been installed from Conda and suggest the appropriate `conda update nextstrain-cli` command.

This may be difficult to get right, as there's quite a bit of nuance. Considerations include:

1. The Conda environment path/prefix named in the `conda update` command (if any) should reflect the environment that the CLI Conda package was _installed into_, not just the top-most activated environment. This comes into play with layered environment activation.
2. An activated Conda environment doesn't mean the CLI is installed into it.
3. The CLI can be installed into a Conda env and run without ever activating the environment.
4. In many cases, the CLI is installed into a Conda environment with Pip instead of a Conda package, in which case the existing upgrade hints should be used.
5. Detection may need to look for a Conda package installation metadata file, e.g. `…/conda/envs/*/conda-meta/nextstrain-cli-*.json`.
6. The structure of Python's `site-packages` dir in a Conda package install is indistinguishable from a non-Conda install.

The following Conda environment variables may be useful, although they do not by themselves indicate that the CLI is installed from Conda, just that some Conda env is active while the CLI is running.

- `CONDA_PREFIX`: Base path of the top-most activated Conda environment.
- `CONDA_DEFAULT_ENV`: Name of the top-most activated Conda environment, if named (not all envs must have names).

It may be useful to introspect [metadata of installed Python distributions](https://packaging.python.org/en/latest/specifications/recording-installed-packages/) when addressing this issue.

tsibley:
Now that we universally recommend install via Conda, this is probably more important to address.

tsibley:
I dug into this a bit this morning. Below are a copy of [my notes](https://github.com/tsibley/blab-standup/blob/master/2022-07-21.md#readme):

A shell with `conda init` seems to have these persistently defined:

    CONDA_SHLVL=0
    CONDA_EXE=/home/tom/.conda/bin/conda
    CONDA_PYTHON_EXE=/home/tom/.conda/bin/python

When running `nextstrain` as installed from the Conda package, it's path (e.g.  `_` or maybe `sys.argv[0]`?) is a cousin of `CONDA_EXE`:

[...]

PR #162, opened 2022-03-28
Document how to upgrade the CLI itself

tsibley:
I reached for a reference to this page in some Groups migration doc I
was writing but the page didn't exist.  Now it does!

### DIFF
--- a/doc/upgrading.rst
+++ b/doc/upgrading.rst
@@ -1,0 +1,45 @@
+=========
+Upgrading
+=========
+
+This page describes how to upgrade the Nextstrain CLI--the ``nextstrain``
+command--itself, without also upgrading the entire Nextstrain
+:term:`docs:runtime`.
+
+The first step is to figure out what kind of Nextstrain CLI installation you
+have.
+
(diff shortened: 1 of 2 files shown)


tsibley:
Page would be simplified by https://github.com/nextstrain/cli/issues/114.

doc/upgrading.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. code-block:: console
+    :emphasize-lines: 6
+
+    $ nextstrain check-setup
+    A new version of nextstrain-cli, X.Y.Z, is available!  You're running A.B.C.
+
+    Upgrade by running:
+
+        [UPGRADE COMMAND]
```
victorlin, line 32: We should somehow add the `mamba update` instructions in this output, otherwise users following the central installation docs will see `pip install` which is not advised for conda-based installations.

tsibley: Yep, see https://github.com/nextstrain/cli/issues/114.

victorlin: Ah ok, missed your comment above.

victorlin (approved):

Issue #96, opened 2020-06-30
[update] Stop overwriting `nextstrain/base:latest` in `~/.nextstrain/config`


joverlee521:
Running `nextstrain update` sets the `docker.image` config to the `nextstrain/base:build-*` tag even when `docker.image` is explicitly set to `nextstrain/base:latest`.

This is due to this bit of code within the [`latest_build_image()` function](https://github.com/nextstrain/cli/blob/master/nextstrain/cli/runner/docker.py):
```python
if tag == "latest" or is_build_tag(tag):
        build_tags = sorted(filter(is_build_tag, tags(repository)))

        if build_tags:
            return repository + ":" + build_tags[-1]
        else:
            return repository + ":latest"
    else:
        return image_name
```

tsibley:
Good make sure with this that on first run, the default image still leads to a pinned `build-*` tag. I think the desired behaviour can be achieved by conditioning on implicit vs. explicit `latest` tag.

PR #163, opened 2022-03-31
runner.docker: Distinguish between an explicit vs. implicit "latest" image tag when updating

tsibley:
Behaviour with an implicit "latest" tag remains the same: it resolves to
the most recent "build-*" tag when `nextstrain update` is run which then
pins it in the config file.

An explicit "latest" tag used to do the same but now is preserved: it's
not updated to the most recent "build-*" tag during `nextstrain update`.
This allows folks to opt into tracking the leading edge of changes to
the image.

Resolves <https://github.com/nextstrain/cli/issues/96>.

### Testing
Tested locally and works for me. Also added additional doctests.

### DIFF
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -17,6 +17,10 @@ from ..volume import store_volume
 from ..__version__ import __version__
 
 
+# The default intentionally omits an explicit "latest" tag so that on the first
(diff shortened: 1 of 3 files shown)


joverlee521 (approved):
Thanks for fixing this! Worked for me 👍

tsibley:
Released in 3.2.4 just now.

Issue #167, opened 2022-04-05
Inconsistent image version usage when no docker.image setting


tsibley:
_This issue is mostly documenting and describing something I noticed in the process of writing #163. I think this issue will mostly be resolved as a consequence of other related work we want to do (primarily https://github.com/nextstrain/cli/issues/166, if we implement the proposed solution there), but still wanted to describe it separately for future reference._

The default image used for the Docker and AWS Batch runners (e.g. when no `--image` option is given) is determined by:

https://github.com/nextstrain/cli/blob/c76c7b8a88706ddb28ea4d7e9ae8de38936c0d7b/nextstrain/cli/runner/docker.py#L20-L22

The environment variable override is mostly meant for automation and development contexts and can be ignored for the purposes of this issue as anyone setting it is doing so explicitly and can be assumed to know what they want. That leaves the `docker.image` config setting and the final fallback default of `nextstrain/base`.

The `docker.image` setting is only set and updated in normal usage when `nextstrain update` is run, where it tracks the latest `nextstrain/base:build-*` image. This leads to a consistent image used for both the local Docker and remote AWS Batch runtimes. If `nextstrain update` has never been run though—we don't include it in our installation instructions for example—then `docker.image` will be missing and the final fallback of `nextstrain/base` will be used. Docker will automatically download the most recent `latest` tag on first `nextstrain build` (or `view`) and then use that image version for all subsequent runs, which is ok. But running on AWS Batch will always track the most recent `latest` tag regardless of the snapshot of `latest` used by local Docker or previous Batch runs.

This inconsistency would be good to iron out:

- [ ] If not present, then set `docker.image` to the latest `build-*` tag during `nextstrain check-setup --set-default`. This means in most cases the user will get a consistent image version across runtimes and some version reporting will be better (because it'll report a specific `build-*` tag instead of a snapshot of `latest`). The inconsistent behaviour still exists if `--set-default` is not used.
- [ ] Ultimately make Docker and AWS Batch more consistent regardless of the `docker.image` setting by implementing the proposed solution to https://github.com/nextstrain/cli/issues/166
- [ ] Advise `nextstrain update` after installation? Or have `check-setup` download the image (i.e. built-in `update`)?  Though this would make the command much slower and requires some additional work to not muddy the output. Maybe not necessary.

Good solutions may be related to/integrated with the other "first time setup" topic of [picking a default runner (#164)](https://github.com/nextstrain/cli/issues/164).

PR #168, opened 2022-04-05
runner.docker: Set docker.image setting to the latest build-* tag by default

tsibley:
This is called when the runner is set as the default runner by
`nextstrain check-setup --set-default`.

Partially resolves <https://github.com/nextstrain/cli/issues/167>.

---

See other commit messages for more details.

### Testing
Tested new setting of defaults locally. `pytest` passes locally.

### DIFF
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -94,18 +94,20 @@ def run(opts: Options) -> int:
 
     # Print overall status.
     supported_runners = [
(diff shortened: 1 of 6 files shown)


victorlin (approved):
Looks like this PR is linked to close #167, just making sure that's intentional.

tsibley:
It's not, but I can't unlink it because of the reference in the commit message. (Well, I could rewrite the commit message, but meh, it's clear to humans.)

PR #172, opened 2022-05-10
Handle invalid token 401 Unauthorized responses from nextstrain.org

tsibley:
Adds handling of particular 401 Unauthorized responses from nextstrain.org that indicate a stale or otherwise invalid token. Also reworks how auth is added to requests so it'll be easier to implement auto-renewal/retry on these failures in the future. See commit messages for details.

This is primarily a UX change in the error message presented when a `nextstrain remote` command fails because of a specific reason:

Before (generic):

```
Error: Permission denied.

Are you logged in as the correct user?  Current user: trs.

If your permissions were recently changed (e.g. new group
membership), it might help to run `nextstrain login --renew`
and then retry your command.
```

After (more specific):

```
Error: Login credentials appear to be out of date.

Please run `nextstrain login --renew` and then retry your command.
```

### Related issue(s)
Related to https://github.com/nextstrain/nextstrain.org/pull/537.

### Testing
Tests pass locally, including new doctests. Tested against local nextstrain.org server running https://github.com/nextstrain/nextstrain.org/pull/537.

### DIFF
--- a/nextstrain/cli/remote/nextstrain_dot_org.py
+++ b/nextstrain/cli/remote/nextstrain_dot_org.py
@@ -61,6 +61,7 @@ import urllib.parse
 from collections import defaultdict
 from email.message import EmailMessage
 from pathlib import Path, PurePosixPath
+from requests.utils import parse_dict_header
 from textwrap import indent
 from typing import Dict, Iterable, List, NamedTuple, NewType, Optional, Tuple, Union
 from urllib.parse import urljoin, urlsplit, quote as urlquote, quote_plus as urlquote_plus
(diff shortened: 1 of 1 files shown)


nextstrain/cli/remote/nextstrain_dot_org.py
```diff
@@ -600,6 +633,37 @@ def raise_for_status(response: requests.Response) -> None:
...
+
+    >>> r = requests.Response()
+    >>> authn_challenge(r) # None
+    """
+    challenge = response.headers.get("WWW-Authenticate")
+
+    if not challenge:
+        return None
+
+    challenge_params = remove_prefix("Bearer ", challenge, ignore_case = True)
```
victorlin, line 658: _[non-blocking]_

Is there a particular reason to ignore case here? I don't know much about OAuth but all the examples I'm seeing use the `Bearer` casing.

victorlin (approved):

nextstrain/cli/remote/nextstrain_dot_org.py
tsibley: Postel's Law… but maybe unnecessarily so. The RFC [mandates `Bearer` literally](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1). I also note when checking related RFCs that challenges (what appear in `WWW-Authenticate` and elsewhere) allow multiple spaces (`1*SP`) between the `auth-scheme` (e.g. `Bearer`) and the params, whereas the code in this branch trims only 1 space and the rest will end up in the name of the first parameter (likely breaking parsing if present).

I'll fix these two things up.

tsibley: Done: https://github.com/nextstrain/cli/compare/949971d3cec6a9334893ddc0e49216c836b3ad16..6a62b60fba7427ffee8f277d0002303928b7b15b

victorlin (approved):

Issue #144, opened 2022-02-04
[build] --cpus and --memory don't work with newer AWS Batch job definitions


tsibley:
The `--cpus` and `--memory` options for `nextstrain build` don't work with newer AWS Batch job definitions that use entries in `resourceRequirements` (instead of separate `vcpus` and `memory` properties) to declare default CPU and memory requirements.

More information on this deprecation is at https://docs.aws.amazon.com/batch/latest/userguide/troubleshooting.html#override-resource-requirements and displayed in a tooltip in the AWS console:

> This value was submitted using containerOverrides.vcpus which has been deprecated and was not used as an override. Instead, the VCPU value found in the job definition's resourceRequirements key was used instead. More information about the deprecated key can be found in the AWS Batch API documentation.

Our initial troubleshooting [happened in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1643931239045979?thread_ts=1643870962.096539&cid=C01LCTT7JNN).

Notably, using the AWS console to modify an existing job definition which uses `vcpus` and `memory` will switch to `resourceRequirements` in the new revision. However, you can still create definitions using `vcpus` and `memory` via the AWS API or CLI, e.g. `aws batch register-job-definition`.

Ideal result would be supporting both kinds of job definitions without much extra effort. Maybe defining overrides in `resourceRequirements` during submission will work with both kinds of definitions? Should test.

tsibley:
> Ideal result would be supporting both kinds of job definitions without much extra effort. Maybe defining overrides in `resourceRequirements` during submission will work with both kinds of definitions? Should test.

This works! I tested with

```diff
diff --git a/nextstrain/cli/runner/aws_batch/jobs.py b/nextstrain/cli/runner/aws_batch/jobs.py
index 81b0c4b..b2a83be 100644
--- a/nextstrain/cli/runner/aws_batch/jobs.py
+++ b/nextstrain/cli/runner/aws_batch/jobs.py
@@ -177,8 +177,10 @@ def submit(name: str,
                 *forwarded_environment(),
                 *[{"name": name, "value": value} for name, value in env.items()]
             ],
-            **({ "vcpus": cpus } if cpus else {}),
-            **({ "memory": memory } if memory else {}),
+            "resourceRequirements": [
+                *([{ "type": "VCPU", "value": str(cpus) }] if cpus else []),
+                *([{ "type": "MEMORY", "value": str(memory) }] if memory else []),
+            ],
             "command": [
[...]

PR #177, opened 2022-05-26
runner.aws_batch: Use containerOverrides.resourceRequirements instead of .vcpu and .memory

tsibley:
Job submissions with resourceRequirements correctly override the
defaults from job definitions that use the newer resourceRequirements
and those that use the older, deprecated vcpu and memory fields.  Job
submissions using the older fields only override job definitions that
also use the older fields; otherwise they're ignored with only an easy
to miss warning in the AWS console to alert you.

Notably, using the AWS console to modify an existing job definition
which uses vcpus and memory will switch the definition to
resourceRequirements automatically in the new revision.  This meant
revising your old job definition in the AWS console could break --cpus
and --memory for your `nextstrain build` invocations.  It also means
that --cpus and --memory would never have worked if your AWS job
definitions were originally created with resourceRequirements.

Resolves <https://github.com/nextstrain/cli/issues/144>.

### Testing
See #144 for how/what I tested.

### DIFF
--- a/nextstrain/cli/runner/aws_batch/jobs.py
+++ b/nextstrain/cli/runner/aws_batch/jobs.py
@@ -177,8 +177,10 @@ def submit(name: str,
                 *forwarded_environment(),
                 *[{"name": name, "value": value} for name, value in env.items()]
             ],
-            **({ "vcpus": cpus } if cpus else {}),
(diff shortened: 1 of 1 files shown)


victorlin (approved):
Didn't test personally, but your testing looks good.

Issue #175, opened 2022-05-25
[build] Snakemake will auto-detect an incorrect number of CPUs on AWS Batch when `--cpus` is not used


tsibley:
### Current Behavior
In the absence of a `--cpus` option provided to `nextstrain build`, the AWS Batch runner uses the job definition's default CPU allocation (`vcpus`) which is not detectable by Snakemake. Instead, Snakemake detects and uses the number of CPUs on the entire Batch node. This only applies when Snakemake is explicitly run with `--cores` (no args) or `--cores all` or equivalent aliases (`-j`, `-j all`, etc), as otherwise the Snakemake version in our Docker runtime defaults to `--cores 1`.

I expect this will cause oversubscription for the actual CPU time available to the Batch job, but it won't always be so because Snakemake is dealing in the number of processes active at one time within the container and the Batch job's CPU limitation is dealing in cumulative CPU time spent in a period (expressed relative to a single pegged CPU in that same period) across the whole container process tree. So I think oversubscription will mostly be an issue when all the Snakemake processes peg their CPUs at the same time (e.g. alignment, tree building).

### Expected behavior
Snakemake is limited to the same number of CPUs that are nominally provisioned by the Batch job definition when `--cpus` isn't provided.

### How to reproduce
Run this Snakefile under various conditions on AWS Batch and observe that `workflow.cores` reports the number of CPUs of the underlying Batch node if you pass `--cores all`.

```python
import os

rule:
    shell: f"""
        echo workflow.cores {workflow.cores}
        echo nproc "$(nproc)"
        echo os.sched_getaffinity {len(os.sched_getaffinity(0))}
        echo os.cpu_count {os.cpu_count()}
        env | grep -i threads || true
    """
```

### Possible solution
1. Explicitly inform Snakemake of the Batch job description's CPU allocation (`vcpus`), since that's what will apply in the absence of `--cpus`. Downside is this breaks boundaries between the `build` command and the Batch runner: the former is what informs Snakemake but the latter is what knows about Batch job definitions.
2. Require `--cpus` for the AWS Batch runner. Downside is this is a breaking change and erodes the runtime portability properties of `nextstrain build`'s interface.

### Additional context
Realized this was happening when implementing TKTK.

Solution possibly related to https://github.com/nextstrain/cli/issues/144. It at least will touch the same code.

tsibley:
[...]

PR #179, opened 2022-05-27
build: Default Snakemake's --cores option to "all"

tsibley:
Applies when our own `--cpus` option isn't provided.  This will allow upgrading of Snakemake in our Docker runtime without inflicting the addition of `--cores` (or equivalent) arguments onto every existing `nextstrain build` invocation.

Snakemake requires the `--cores` option as of 5.11.¹  This has spawned much discussion² where it was made clear this is an intentional, permanent change and a default will not be added.  Add our own default instead to insulate our users from this and make Nextstrain builds fast-by-default.  Our `--cpus` option can be used to limit CPU usage back from this default if necessary, and users can always specify `--cores` (or equivalents) themselves.

Note that we don't default the `--cpus` option (opts.cpus) because we don't want to _limit_ the CPUs available in the Docker or AWS Batch runtimes, just inform Snakemake that it can use whatever CPUs are available.³  This means that `nextstrain build` will parallelize as much as it can regardless of the runner, which feels right in the context of launching a multi-step workflow.  Note that this is different than Augur's default of using a single CPU, but the context in which Augur is invoked is typically a single-step in a larger workflow where more explicit coordination between steps is required for parallelism.

¹ https://github.com/snakemake/snakemake/issues/283

² https://github.com/snakemake/snakemake/issues/308, https://github.com/snakemake/snakemake/issues/312, https://github.com/snakemake/snakemake/issues/450, https://github.com/snakemake/snakemake/issues/885

³ But there's a caveat here for AWS Batch, see <https://github.com/nextstrain/cli/issues/175>.

### Testing
- [x] Manual testing of permutations with and without `--cpus` and explicit `--cores`, looking at what Snakemake reports for `workflow.cores`
- [x] CI passes

### DIFF
--- a/nextstrain/cli/command/build.py
+++ b/nextstrain/cli/command/build.py
@@ -60,7 +60,9 @@ def register_parser(subparser):
         help    = "Number of CPUs/cores/threads/jobs to utilize at once.  "
                   "Limits containerized (Docker, AWS Batch) builds to this amount.  "
                   "Informs Snakemake's resource scheduler when applicable.  "
-                  "Informs the AWS Batch instance size selection.",
+                  "Informs the AWS Batch instance size selection.  "
+                  "By default, no constraints are placed on how many CPUs are used by a build; "
+                  "builds may use all that are available if they're able to.",
         metavar = "<count>",
         type    = int)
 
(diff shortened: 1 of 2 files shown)


tsibley:
@huddlej I'd be particularly interested in your thoughts here on defaulting to `all`.

nextstrain/cli/command/build.py
```diff
@@ -60,7 +60,9 @@ def register_parser(subparser):
         help    = "Number of CPUs/cores/threads/jobs to utilize at once.  "
                   "Limits containerized (Docker, AWS Batch) builds to this amount.  "
                   "Informs Snakemake's resource scheduler when applicable.  "
-                  "Informs the AWS Batch instance size selection.",
+                  "Informs the AWS Batch instance size selection.  "
+                  "By default, no constraints are placed on how many CPUs are used by a build;"
```
huddlej, line 64: Nitpick, but a space at the end of this line would change the resulting text from:

> By default, no constraints are placed on how many CPUs are used by a build;builds may use all that are

to:

> By default, no constraints are placed on how many CPUs are used by a build; builds may use all that are

huddlej (approved):
Fully onboard with this change!

>  Note that this is different than Augur's default of using a single CPU, but the context in which Augur is invoked is typically a single-step in a larger workflow where more explicit coordination between steps is required for parallelism.

This is a great point. When you're running a workflow, you're concerned about parallelism _and_ multiprocessing, but when you're running a single command (Augur or any other), you're concerned about multiprocessing in the context of other processes you have running. I think I would be frustrated more often by a default of 1 CPU for my workflow than a default of all CPUs.

tsibley:
Thanks @huddlej! I'll rebase this onto current `master` branch and add a changelog entry before merging.

PR #180, opened 2022-06-06
CI build then test + refactoring

tsibley:
See commit messages.

### Testing
- [x] CI passes ;-) (…after much trial and error with remote debugging)

### DIFF
--- a/.github/actions/setup-integration-tests/action.yaml
+++ b/.github/actions/setup-integration-tests/action.yaml
@@ -1,0 +1,58 @@
+name: Setup integration tests
+description: >-
+  Sets up prerequisites for run-integration-tests, namely installing software
+  needed by the "native" runner and cloning zika-tutorial for use as an test
+  build.
+
+  Jobs which use this action must also set
+
+    defaults:
+      run:
+        shell: bash -l -eo pipefail {0}
+
+  at the job level (or the equivalent "shell:" key at the step level) to
+  activate the integration Conda environment by default.
+
+inputs:
+  python-version:
+    description: Version of Python to use for conda-incubator/setup-miniconda.
+    type: string
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: conda-incubator/setup-miniconda@v2
+      with:
+        python-version: ${{ inputs.python-version }}
+        mamba-version: "*"
+        channels: conda-forge,bioconda
+        channel-priority: true
+        activate-environment: test
+
+    - run: cat ~/.profile || true
+      shell: bash -l -eo pipefail {0}
+
+    - run: cat ~/.bashrc || true
+      shell: bash -l -eo pipefail {0}
+
+    # Install software for the "native" runner.
+    - run: mamba install augur auspice snakemake
+      shell: bash -l -eo pipefail {0}
+
+    - run: conda info
+      shell: bash -l -eo pipefail {0}
+
+    - run: conda list
+      shell: bash -l -eo pipefail {0}
+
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -19,63 +19,77 @@ jobs:
...
-          # Assert that we're on the expected Python version, in case the GH
-          # Actions environment is messed up.
-          python3 --version | grep -F 'Python ${{ matrix.python }}.'
+  build-dist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
```
victorlin, line 40: Could this be kept consistent with `*.matrix.python` and the version used in the `doc` job? More concretely, 2 suggestions:

1. Add `3.10` to `test-source.matrix.python`, `test-dist.matrix.python`
2. Update `doc` job to use `3.10`

.github/actions/setup-integration-tests/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+        python --version
+        type python3
+        python3 --version
+        python --version | grep -F 'Python ${{ inputs.python-version }}.'
+        python3 --version | grep -F 'Python ${{ inputs.python-version }}.'
+        [[ "$(python --version)" == "$(python3 --version)" ]]
+
+    # Clone the small build we'll use as an integration test case.
+    - run: git clone https://github.com/nextstrain/zika-tutorial
+      shell: bash -l -eo pipefail {0}
```
victorlin, line 64: _(non-blocking comment)_

[No higher-level `runs.shell`](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-composite-actions), bummer...

victorlin (changes requested):

.github/actions/setup-integration-tests/action.yaml
tsibley: Yeah, I was disappointed.

.github/workflows/ci.yaml
tsibley: I'll add 3.10 to the two version matrices. ISTR some issue with setup-miniconda and 3.10, but it's a very vague recollection so we'll find out.

The doc job uses 3.6 because that's what `.readthedocs.yml` specifies. I suppose I can bump that up too and then bump the `doc` CI job.

tsibley: Ok, a bit of turbulence to get thru.

- Conda indeed has problems with 3.10, reporting: `package augur-10.0.0-py_0 requires bcbio-gff >=0.6.0, but none of the providers can be installed`. I'm going to drop the `test-dist` jobs on 3.10 because of this, but keep the `test-source` jobs.
- [RTD complains](https://readthedocs.org/projects/nextstrain-cli/builds/17121136/) about 3.10 saying that it's not a supported version. Highest is 3.8.
- However, the [config spec now suggests a different field](https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version) to use for setting up Python, so I'll look into that.

tsibley: https://github.com/nextstrain/cli/pull/180/commits/02af3a1839fd10c09e3c79594caeb9b3f0550fc8

https://github.com/nextstrain/cli/pull/180/commits/d12f5b5bfe65e527bdf2cf050c06dc8fcc6295ac

victorlin (approved):

Issue #31, opened 2018-10-02
Windows support


tsibley:
I would like the CLI to work well on Windows when used with Docker for Windows on Windows 10 Professional or Enterprise in a very vanilla setup. I think it is most of the way there, but there are some showstoppers and several rough edges and sharp corners. It would also be very nice, and serve more users users, to support Docker Toolbox on Windows 10 Home or older versions of Windows.  To do any of this requires access to a Windows machine (not a VM, unfortunately) for development and testing.

This issue is going to serve as a punch list of things that need addressing. Some items may be moved to separate issues in the future.

- [ ] Python 3 installs just as `python`, no `python3`.
- [ ] netifaces wheel must be available for install. The current version is missing for Python 3.7, but this will be present in the next version (thanks to [my PR](https://github.com/al45tair/netifaces/pull/19)).
- [ ] Paths for `docker run` invocation must not be standard Windows paths, but munged versions. Affects volume mapping.
- [ ] Unicode not well-supported it seems in cmd.exe or PowerShell
- [ ] Terminal escapes may not be well-supported as-is
- [x] os.execvp() doesn't work as expected

tsibley:
Relatedly, there's been some good news for Linux environments on Windows, including Docker support within WSL (alleviating a lot of problems, including making the Pro vs. Home edition issue moot): https://devblogs.microsoft.com/commandline/announcing-wsl-2/

victorlin:
>  netifaces wheel must be available for install. The current version is missing for Python 3.7, but this will be present in the next version (thanks to [my PR](https://github.com/al45tair/netifaces/pull/19)).

@tsibley this is an issue again with Python 3.9:

- The last commit to `netifaces` was on 2021-05-31, before AppVeyor added support for Python 3.9.
- The author of `netifaces` has stopped maintaining it and archived the repo, so PR to fix is no longer an option.

Is there an alternative to `netifaces`?

tsibley:
Hmm. There wasn't a viable `netifaces` alternative back in 2018. I suppose there might be nowadays, but I wouldn't be surprised if not.

Some options I see (not exhaustive, obviously):

1. [Offer to maintain `netifaces`](https://github.com/al45tair/netifaces/issues/78) at a bare minimum level of occasionally updating CI to build on new Python versions. This seems like only marginal effort above submitting a PR to an active maintainer. There don't currently seem to be [any forks doing this work already](https://github.com/al45tair/netifaces/network).
[...]

PR #181, opened 2022-06-06
CI on Windows

tsibley:
Adds Windows as a platform on which we run our CI tests and fixes the issues that arise. See commit messages for details.

### Related issue(s)
Related to #14 and #31.

Built on top of #180.

### Testing
- [x] CI passes (after much iteration)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
         python:
           - '3.6'
           - '3.7'
@@ -65,6 +66,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
         python:
           - '3.6'
           - '3.7'
--- a/.github/actions/run-integration-tests/action.yaml
+++ b/.github/actions/run-integration-tests/action.yaml
@@ -24,14 +24,15 @@ runs:
 
         nextstrain version --verbose
 
-    - if: runner.os != 'macOS'
+    - if: runner.os != 'macOS' && runner.os != 'Windows'
       name: Build zika-tutorial with --docker
       shell: bash -l -eo pipefail {0}
       run: |
         git -C zika-tutorial clean -dfqx
         nextstrain build --docker --cpus 2 zika-tutorial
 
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
```
victorlin, line 16: image

Dang, this now translates to 52 jobs per push to a non-fork PR. Did this take ~30m for everything to finish? Sounds like more reason to address nextstrain/private#48.

victorlin (approved):
Looks good!

.github/workflows/ci.yaml
tsibley: Not sure exactly how long it takes given it depends on parallelism and runner availability (more acute with macOS), but yeah, it definitely takes longer now. :grimacing: That said, I think the most likely to fail tests run in the first few minutes, so that's should be an early indicator.

One quick change that I'd be inclined to make is to only run on `push` for `master` and otherwise only run on `pull_request`. This would slash total checks/jobs from 53 (52 + 1) → 27 (26 + 1).

tsibley: (The biggest downside of that is you could no longer push without opening a PR just to see what CI says, but I think that's an acceptable tradeoff, esp. with draft PRs now.)

tsibley:
I'll wait to merge this until #180 is in.

.github/workflows/ci.yaml
victorlin: I'm in favor of that change. Especially since the limit is org-level, it'd be silly to have CI here block arguably more important things such as the [now-hourly monkeypox-ingest](https://github.com/nextstrain/monkeypox/blob/9c3f0b0faac63a94fd726c9174ed3456d78388b7/.github/workflows/fetch-and-ingest.yaml#L19).

tsibley: Sounds good. I'll add it to #180 since I'm adding the 3.10 stuff there anyway.

tsibley: Added there as https://github.com/nextstrain/cli/pull/180/commits/fab709a2f45fc74afe2a15038e877e4dd58ae222

Issue #185, opened 2022-06-11
Licensing of standalone installation archives


tsibley:
_This is a stub issue to remind myself to come back to this._

The [license report from a recent standalone build](https://gist.github.com/tsibley/3ae783acff14ad5924a97847fa1b7da5) is a super informative summary. I'm very impressed by PyOxidizer automatically doing this! See also the `COPYING.txt` file included in each standalone installation archive for the full blow-by-blow.

Notably, all the unknown licenses reported are the result of package metadata that doesn't use SPDX license identifiers and none of the unknown licenses are "new" licenses for the bundle, so we can just look at the nice license tally matrix at the top.

There we see that the only copyleft license (which means it potentially carries provisions that we might not be able to meet when bundling) is the MPL 2.0, which permits "a larger work using the licensed work [to] be distributed under different terms and without source code for files added in the larger work" per https://choosealicense.com/licenses/mpl-2.0/. Rad.

While doing this manually once in a while, particularly when we add/remove deps is ~ok, a tool like [pip-license](https://github.com/raimon49/pip-licenses) could help us formalize this process and make it part of CI. Once we "set it", we could then "forget it" until a new license was flagged.

PR #186, opened 2022-06-11
Release to PyPI and GitHub Releases from CI

tsibley:
Simplifies the local release process and builds upon other CI work to
make the dists built in CI the basis for everything downstream (tests,
standalone builds, and now PyPI too).  This also means we can include in
GitHub Releases the standalone installation archives, so they're not
squirreled away only in ephemeral Actions artifacts storage.

### Testing
Real test will come with the first releases after this merges, but still, I:

- [x] Tested the release process in a personal fork, with only minor modifications
- [x] Tested all the changelog extraction stuff manually

### DIFF
--- a/devel/changes
+++ b/devel/changes
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+"""
+Extract changelog for a specified version (or __NEXT__ if no version
+specified).
+"""
+from pathlib import Path
+from sys import argv, exit, stdout, stderr
+
+
+CHANGELOG = Path(__file__).parent / "../CHANGES.md"
+
+
+def main(version = "__NEXT__"):
+    with CHANGELOG.open(encoding = "utf-8") as file:
+        for line in file:
+            # Find the heading for this version
+            if line.rstrip("\n") == f"# {version}" or line.startswith(f"# {version} "):
+
+                # Print subsequent lines until we reach the next version heading
+                for line in file:
+                    if line.startswith("# "):
+                        break
+                    stdout.write(line)
+                return 0
+
+        print(f"No changes found for {version!r}.", file = stderr)
+        return 1
+
+
+if __name__ == "__main__":
+    exit(main(*argv[1:]))
--- a/devel/create-github-release
+++ b/devel/create-github-release
@@ -1,0 +1,70 @@
+#!/bin/bash
+set -euo pipefail
+shopt -s extglob
+
+devel="$(dirname "$0")"
+
+main() {
+    local version="${1:?version is required}"
+    local commit pre_release
+    shift
+    local -a assets=("$@")
+
+    if [[ ${#assets[@]} -eq 0 ]]; then
+        echo "ERROR: no assets provided" >&2
+        exit 1
+    fi
+
+    # Asserts that $version is an actual tag, not a branch name or other ref.
+    #
+    # Prefixing with refs/tags/ is a bit hacky and suffixing with ^{tag} would
+    # be more proper to restrict to a proper annotated, meant-for-release tag.
+    # While this is dandy in any given clone, in the GitHub Actions context the
+    # annotated tag is intentionally **overwritten** and converted to a
+    # lightweight tag by actions/checkout@v3, which doesn't/can't pass the
+    # muster of ^{tag}.
+    git rev-parse --verify "refs/tags/$version" >/dev/null
+
+    # Translate from tag into commit for `gh release create --target`.
+    commit="$(git rev-parse --verify "$version^{commit}")"
+
+    if is-pre-release "$version"; then
+        pre_release=1
+    fi
+
+    gh release create \
+        "$version" \
+        --repo "${GITHUB_REPOSITORY:-nextstrain/cli}" \
+        --title "$version" \
+        --target "$commit" \
+        ${pre_release:+--prerelease} \
+        --notes-file <(preamble; "$devel"/changes "$version") \
+        "${assets[@]}"
+}
+
+is-pre-release() {
+    # See https://peps.python.org/pep-0440/
+    local version="$1"
+    case "$version" in
+        *@(a|b|rc|c)+([0-9])*)  # alpha, beta, release candidate (PEP 440 pre-releases)
+            return 0;;
+        *.dev+([0-9])*)         # dev release
+            return 0;;
+        *+*)                    # local version
+            return 0;;
+        *)
+            return 1;;
+    esac
+}
+
+preamble() {
+    cat <<~~
+
+_These release notes are automatically extracted from the full [changelog][]._
+
+[changelog]: https://github.com/nextstrain/cli/blob/master/CHANGES.md#readme
+
+~~
+}
+
+main "$@"
--- a/devel/create-github-release
+++ b/devel/create-github-release
@@ -16,14 +16,7 @@ main() {
     fi
 
     # Asserts that $version is an actual tag, not a branch name or other ref.
-    #
-    # Prefixing with refs/tags/ is a bit hacky and suffixing with ^{tag} would
-    # be more proper to restrict to a proper annotated, meant-for-release tag.
-    # While this is dandy in any given clone, in the GitHub Actions context the
-    # annotated tag is intentionally **overwritten** and converted to a
-    # lightweight tag by actions/checkout@v3, which doesn't/can't pass the
-    # muster of ^{tag}.
-    git rev-parse --verify "refs/tags/$version" >/dev/null
+    git rev-parse --verify "$version^{tag}" >/dev/null
 
     # Translate from tag into commit for `gh release create --target`.
     commit="$(git rev-parse --verify "$version^{commit}")"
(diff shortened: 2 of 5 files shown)


tsibley:
One thing I'll note is that I'm ok including the standalone installation archives in the GitHub Release because of [my initial review of licenses for the software we bundle](https://github.com/nextstrain/cli/issues/185). I still don't think they're ready for publicizing yet, but I don't think it'll hurt to quietly link to them in Releases for now, and it'll help us get other tooling in order.

tsibley:
Additionally, while this preserves the local creation of release tags with `./devel/release`, if desired we can (in another PR) make a "Tag a release" GitHub Actions workflow which just runs `./devel/release` for you. I don't anticipate personally wanting/needing this, but it's an option available because the CI release process is based on the tag and doesn't care where it comes from.

devel/create-github-release
```diff
@@ -0,0 +1,70 @@
...
+
+    if is-pre-release "$version"; then
+        pre_release=1
+    fi
+
+    gh release create \
+        "$version" \
+        --repo "${GITHUB_REPOSITORY:-nextstrain/cli}" \
+        --title "$version" \
+        --target "$commit" \
```
victorlin, line 39: Wait, is `--target` even necessary since `"$version"` is already provided? I tried running locally without it and seems to work fine.

devel/changes
```diff
@@ -0,0 +1,11 @@
...
+# Extract changelog for a specified version (or __NEXT__ if no version
+# specified).
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+version="${1:-__NEXT__}"
+export version
+
+perl -ne 'print if /^# \Q$ENV{version}\E( |$)/ ... (/^# / and exit)' CHANGES.md | tail -n +2
```
victorlin, line 11: _(non-blocking)_

Could you break this down to a perl newbie? So far I understand:

- `-n`: loop through lines in CHANGES.md
- `-e`: execute a line of perl
- `^# `: match start of string + `# `.
- `\Q$ENV{version}\E`: match the version exactly (escape chars from regex)
- `( |$)`: match ` ` or EOL

devel/create-github-release
tsibley: From [the `gh` docs](https://cli.github.com/manual/gh_release_create):

> If a matching git tag does not yet exist, one will automatically get created from the latest state of the default branch. Use `--target` to override this.

devel/changes
tsibley: I break it down below because learning is fun… but also maybe I shouldn't have just gone with my semi-cheeky example from another PR review comment and instead written a bit more verbose version in a language the team knows. :upside_down_face: :sweat_smile: Perhaps [one of these variants](https://gist.github.com/tsibley/3db7f8df1ec61d407354b0e6c36ba170) (probably _c_), for example.

Break down of:

    perl -ne 'print if /^# \Q$ENV{version}\E( |$)/ ... (/^# / and exit)' CHANGES.md

This is a Perl "one-liner" which loops through lines (`-n`) of input evaluating (`-e`) a bit of code for each line.  (Very much like `sed` or `awk`.)  The current line of input is implicitly assigned to Perl's global "topic variable", `$_`, which can be thought of as "the current thing the code is talking about".  For many functions, the topic is the default argument if no explicit argument is provided.

Input is from the file `CHANGES.md` in this case, but the command would read from stdin if no filenames were passed (or multiple files sequentially if multiple filenames were passed).

The code conditions a call to the core `print` function (a keyword) with a postfix `if` statement.  The `print` function when given no arguments will use the topic variable (i.e. the current line of input here).  Postfix `if` statements are identical to prefix `if` statements and are only syntactic sugar.

The `if`'s condition uses the [triple-dot range operator](https://perldoc.perl.org/perlop#Range-Operators), `...`, which maintains a boolean state based on its left-hand side (LHS) and right-hand side (RHS).  It evalutes to false initially, turns true when the LHS evalutes to true, and stays true until the RHS evalutes to false, at which point it turns false again (until the LHS turns true again, if ever).  (The range operator has origins in `sed` and `awk`, as does the idea of a topic line, which Perl extends to the topic variable more generally.)

The LHS of the range operator tests the topic against a regular expression, which you already grok.

The RHS of the range operator tests a compound expression which evalutes to false unless the topic matches the regular expression, in which case it exits the program early.

It might be illustrative to see an equivalent program increasingly "spelled out":

```perl
while (<ARGV>) {
    print if /^# \Q$ENV{version}\E( |$)/ ... (/^# / and exit);
}
```

```perl
my $found = 0;

while (<ARGV>) {
    if ($found) {
        exit if /^# /;
        print;
    }
    elsif (/^# \Q$ENV{version}\E( |$)/) {
        $found = 1;
        print;
    }
}
```

Note that this still uses the topic variable.  It too could be "spelled out" with explicit lexical variables, but the topic is an idiomatic part of Perl in many contexts.

```perl
my $found = 0;

while (my $line = <ARGV>) {
    if ($found) {
        exit if $line =~ /^# /;
        print $line;
    }
    elsif ($line =~ /^# \Q$ENV{version}\E( |$)/) {
        $found = 1;
        print $line;
    }
}
```

Ok, I'll stop there, as this is likely to be more than you wanted to know about Perl.

devel/changes
```diff
@@ -0,0 +1,29 @@
...
+
+
+CHANGELOG = Path(__file__).parent / "../CHANGES.md"
+
+
+def main(version = "__NEXT__"):
+    with CHANGELOG.open(encoding = "utf-8") as file:
+        for line in file:
+            # Find the heading for this version
+            if line.startswith(f"# {version}") or line.startswith(f"# {version} "):
```
victorlin, line 17: _(non-blocking)_

If `line.startswith(f"# {version}")` is true then `line.startswith(f"# {version} ")` will also be true, so the latter is not needed?

```suggestion
            if line.startswith(f"# {version}"):
```

devel/create-github-release
```diff
@@ -0,0 +1,70 @@
...
+
+    # Asserts that $version is an actual tag, not a branch name or other ref.
+    #
+    # Prefixing with refs/tags/ is a bit hacky and suffixing with ^{tag} would
+    # be more proper to restrict to a proper annotated, meant-for-release tag.
+    # While this is dandy in any given clone, in the GitHub Actions context the
+    # annotated tag is intentionally **overwritten** and converted to a
+    # lightweight tag by actions/checkout@v3, which doesn't/can't pass the
+    # muster of ^{tag}.
+    git rev-parse --verify "refs/tags/$version" >/dev/null
```
victorlin, line 26: _(non-blocking)_

You probably already saw this, but `git fetch --tags --force` from https://github.com/actions/checkout/issues/290#issuecomment-680260080 is another option:

```suggestion
    # Asserts that $version is an actual tag, not a branch name or other ref.
    #
    # In the GitHub Actions context the annotated tag is intentionally
    # **overwritten** and converted to a lightweight tag by
    # actions/checkout@v3. Overwrite the tags in original form so ^{tag}
    # (which only works on annotated tags) can be used properly.
    git fetch --tags --force
    git rev-parse --verify "$version^{tag}" >/dev/null
```

devel/create-github-release
victorlin: I see. Would be nice to have `--target` accept a tag or `refs/tags/<tag>` but seems like that's not currently supported (https://github.com/cli/cli/issues/2133#issuecomment-1164899636).

victorlin (approved):
Looks good, just a couple small non-blocking things. Tested on a personal fork.

devel/changes
tsibley: Ah, I got this wrong. Thanks for spotting it. It should be.

```suggestion
             if line == f"# {version}" or line.startswith(f"# {version} "): 
```

(The goal is to avoid matching `# 1.2.10` when the version is `1.2.1` and account for `# __NEXT__` not having a trailing space.)

tsibley: Fixed.

devel/create-github-release
tsibley: Ah, I actually hadn't considered manually fetching the actual annotated tag object, but that's maybe a better idea. I would want to do that inside the workflow though, not inside this program, to properly scope the solution to the context in which it occurs (and ensure that if the program is run locally it doesn't muck around in your clone). I'd probably also not use the hammer of `git fetch --tags --force` but instead `git fetch --force origin "$GITHUB_REF"` or `git fetch --force origin tag "$GITHUB_REF_NAME"`.

devel/create-github-release
tsibley: Agreed! I'd tried with a tag but it didn't work, and was unsurprised since the doc says it must be branch name or full commit id. On the other hand, the error message from the API refers to the `target_commitish` field, and a tag is a commit-ish in git parlance (see `man gitglossary`), so from that name you'd expect it to work!

devel/create-github-release
tsibley: Implemented as 203adee.

PR #193, opened 2022-07-11
dev: Strip __NEXT__ from the CHANGES in a release

tsibley:
It's nicer to have a cleaner document in the release artifacts and tag,
while keeping __NEXT__ there as the placeholder for development.

### Testing
- [x] Ran `./devel/release` locally and observed that the __NEXT__ references were removed in the release commit/tag and present again in the dev version bump.


victorlin (approved):
👌

PR #194, opened 2022-07-11
Fix suggested upgrade instructions for standalone installations

tsibley:
See commit messages for details, but the main change here is fixing the suggested upgrade instructions for standalone installations when a new version is detected by `nextstrain update` or `check-setup`. Probably no one but me is using the standalone installs yet so this is a minor thing… but the reason _I'm_ doing so is to notice bugs just like this. :-)

### Related issue(s)
Related to #182.

### Testing
- [x] CI passes
- [x] Manual testing with `update` and the new `NEXTSTRAIN_CLI_LATEST_VERSION` override. Would love cram tests for this sort of thing—I have a prototype—but not pursuing those for now.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,9 @@ development source code and as such may not be routinely kept up to date.
 
 ## Improvements
 
+* The new version check now links out to the changelog for the latest version
+  so you know what you're gonna get.
+
 * The new version check now detects standalone installations and provides
   correct upgrade instructions.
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -13,6 +13,12 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Development
+
+* A new command, `debugger`, was added as a tool to help with troubleshooting
+  environment and execution context issues.  The only thing it does is invoke
+  pdb from within the command's context.
```
victorlin, line 33: I haven't used pdb but will note to try it next time I need to debug this project!

victorlin (approved):

PR #197, opened 2022-07-13
dev: Enable a wider range of flake8 correctness checks

tsibley:
While looking for a flake8 check to catch import shadowing (but alas not
finding what I wanted), I realized that the whole class of "F" checks
from pyflakes are worthwhile correctness checks (in contrast to the "E"
checks, which are mostly stylistic).

Enabling all "F" checks mostly catches a slew of unused imports that
have snuck in over the years.  I would argue these are more correctness
and readability checks than style checks.

### Testing
- [x] CI passes
- [x] Tests pass locally
- [x] Manually exercised functionality of commands implicated by diff

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,9 +4,7 @@
 [flake8]
 select =
     E9,     # syntax errors
-    F63,    # syntax warnings (e.g., "==" versus "is")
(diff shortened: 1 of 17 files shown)


victorlin (approved):
💯

PR #198, opened 2022-07-18
docs: Enable the copybutton extension we enable with our Sphinx theme

tsibley:
Our theme enables and configures the copybutton extension, but this only
works if our theme is also enabled as an extension (not just a theme).

### Related issue(s)
Related to https://github.com/nextstrain/sphinx-theme/pull/26.

### Testing
- [ ] CI passes
- [ ] Copy button in RTD preview build


victorlin (approved):
Good catch!

PR #209, opened 2022-07-27
remote.nextstrain_dot_org: Yield resource URL before deletion not after

tsibley:
A pre-action yield matches the messaging of the `remote delete` command
as of "remote delete: Update messaging to match pre-deletion behaviour"
(bb4ab4c) and the behaviour of the S3 remote as of "remote.s3: Rework
delete() to have a similar API to the other commands" (cc86a1a).

The post-action yield was an oversight and bug in the initial remote
implementation.

### Testing
- [x] Ran updated command locally
- [ ] CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,14 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Bug fixes
+
+* Using `remote delete` on nextstrain.org now correctly outputs the "Deleting…"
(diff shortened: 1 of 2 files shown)


tsibley:
Whoops, I forgot #210 was based on this one and merged it first, thus merging this. I'm ok with that, given the scope of this change. My thinking for splitting them is that this obvious fix could get merged quickly and the `--dry-run` stuff might take more time, but it didn't!

Issue #149, opened 2022-02-11
Add --prompt and --dry-run modes for `remote` family of commands


tsibley:
It would be useful to have `--prompt` and `--dry-run` modes for each of the `nextstrain remote` commands (excepting `ls`).

`--dry-run` mode would go through the motions of doing the thing, as much as possible, but doesn't actually do the thing. This lets users see what _would_ happen if they remove the `--dry-run` option and re-run, e.g. see how dataset paths would be downloaded, uploaded, or deleted. Dry runs of `list` (`ls`) don't make sense and aren't necessary.

`--prompt` mode would be similar, but instead of not doing the thing, it would stop and ask the user to confirm before proceeding. This would reasonable as a default, although that's up for discussion.

`--no-prompt` would negate any previous `--prompt` (explicitly provided or implicit if default), but not `--dry-run`.

An alternative way to think about it is that `--dry-run` is like `--prompt` + an automatic answer of "no" or "cancel". These exact same modes exist in the work I did for [id3c](https://github.com/seattleflu/id3c)'s transactional database commands.

tsibley:
It may be difficult given current command implementations to make `--prompt` work with a single question instead of asking for each action (quickly overwhelming, more state to track). If that's the case, I may implement only `--dry-run`, which is still useful on its own.

One sorta-workaround even given the current design is to implement `--dry-run` and then implement `--prompt` by first running internals with `--dry-run`, then prompting, and then re-running internals without `--dry-run`. The big caveat here is that given any change in local filesystem or remote server state, the wet run could end up doing something different than the dry run.

tsibley:
(That caveat of course applies to the normal use of `--dry-run` by a user to preview changes and then run without `--dry-run`, but it feels more "wrong" to codify it into how `--prompt` works.)

PR #210, opened 2022-07-27
remote: Add --dry-run modes to download, upload, and delete commands

tsibley:
This mode, as is broader convention, goes through the motions of doing
the thing, as much as possible, but doesn't _actually_ do the thing.
That is, no changes should occur when --dry-run is active.

This is particularly useful for seeing what will happen if you're unsure
of how a path or argument is handled.  Dry runs of the list (ls) command
don't make sense and aren't included.

Resolves <https://github.com/nextstrain/cli/issues/149>.

### Testing
- [x] Ran updated commands locally with and without `--dry-run` against both nextstrain.org and S3 remotes
- [x] Tests pass locally
- [x] CI passes


joverlee521 (approved):
LGTM.

I was briefly confused why the `nextstrain_dot_org` side didn't an equivalent `purge_cloudfront` function, then remembered those API endpoints are currently for groups only.

tsibley:
> I was briefly confused why the `nextstrain_dot_org` side didn't an equivalent `purge_cloudfront` function, then remembered those API endpoints are currently for groups only.

And even when they're not for Groups only, the CloudFront purging would also happen server-side not client-side. But also, I've [already moved us to a place](https://github.com/nextstrain/nextstrain.org/pull/552) where we don't need the invalidations for nextstrain.org itself to operate on the latest data. :upside_down_face:

Issue #110, opened 2021-02-22
Allow explicitly overriding the automatic runner selection of `check-setup --set-default`


jameshadfield:
Currently `nextstrain check-setup --set-default` sets the default runner to the first runner which looks like it should work. Currently this list (in order) is docker, native, aws. I'd like this arg to take an optional argument so that the user can set their desired default runner, thus avoiding having to remember to specify it for each `nextrain build` command.

[E.g.](https://github.com/nextstrain/cli/blob/master/nextstrain/cli/command/check_setup.py#L105-L108)
```py
if opts.set_default:
    if opts.set_default==True: # might be opts.set_default=="", haven't tested
        config.set("core", "runner", supported_runners[0])
    elif opts.set_default in runner_status.keys():
        config.set("core", "runner", opts.set_default)
    else:
        print("Error...")
```

Apologies if this has been discussed elsewhere or is a deliberate design decision.

tsibley:
Allowing the user to set a specific default runner is definitely part of the plan, e.g. with something like  `nextstrain config core.runner native` for example. The `config` command doesn't currently exist, but it's on the todo list and there's a draft PR that still needs some refinement.

I hadn't thought of allowing a value to `check-config --set-default`, though. This could work too, in addition to any more general `config` command. Since it's part of `check-setup`, what would the expected behaviour be if the requested default runner is known/valid but not detected as supported? I could see either:

1. Warn, but set as default.
2. Error, don't set as default.

I'd be inclined to (2) since this is part of `check-setup` and thus it would act as an assertion. (`config` could still be used to set it regardless of detected support.)

Thoughts?

In the meantime, you could avoid remembering to specify it every time by editing `~/.nextstrain/config` manually (if you're in a persistent environment) or with a command like:

```
echo -e '[core]\nrunner = native\n' >> ~/.nextstrain/config
```

Although this is not very user-friendly for instructions, which is what I suspect you're after…

(I also thought there was a `NEXTSTRAIN_RUNNER` environment variable which could be set, but I see that there's not.)

jameshadfield:
`nextstrain config core.runner native` seems the best solution here, but I'm unsure of an ETA here (e.g. see https://github.com/nextstrain/cli/issues/68#issuecomment-614161668). While that's in progress, manual editing of `~/.nextstrain/config` would be ok - I forgot this was an option. I'll change this issue to ask for documentation of this. 

[...]

PR #216, opened 2022-08-25
check-setup: Allow --set-default to optionally take a choice

tsibley:
Providing an explicit choice overrides the automatic selection of the
first passing runner.  If the chosen runner isn't supported, it's still
set as the default, but the command exits with an error.  This is akin
to when check-setup run without --set-default and the already-configured
default is unsupported.

Resolves <https://github.com/nextstrain/cli/issues/110>.

### Testing
- [x] Manually tested `--set-default` with and without options under various conditions
- [x] Local tests pass
- [x] CI passes

### DIFF
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -22,7 +22,7 @@ from textwrap import indent
 from .. import config
 from ..types import Options
 from ..util import colored, check_for_new_version, remove_prefix, runner_name
-from ..runner import all_runners, default_runner # noqa: F401 (it's wrong; we use it in run())
+from ..runner import all_runners, all_runners_by_name, default_runner # noqa: F401 (it's wrong; we use it in run())
 
 
 def register_parser(subparser):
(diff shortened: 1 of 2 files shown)


nextstrain/cli/command/check_setup.py
```diff
@@ -103,16 +106,26 @@ def run(opts: Options) -> int:
...
+            set_default = supported_runners[0]
+    elif opts.set_default:
+        set_default = all_runners_by_name[opts.set_default]
+
+    if set_default:
+        default_runner = set_default
+        print()
+        print("Setting default environment to %s." % runner_name(default_runner))
+        config.set("core", "runner", runner_name(default_runner))
+        default_runner.set_default_config()
```
victorlin, line 126: Output is a bit confusing in this scenario:

```sh
# Docker daemon is not running
nextstrain check-setup --set-default docker
```

Output:

```
# docker is not supported

...

Supported Nextstrain environments: native, aws-batch

Setting default environment to docker.

No good.  Default environment (docker) is not supported.
```

Docker gets set as the default even though it's not supported. I think the behavior is fine since it's explicitly requested by the user, but the message `No good.  Default environment (docker) is not supported.` makes it seem like the command failed.

Suggestion: replace `No good.` with `WARNING:`. This output would read better to me:

```
# docker is not supported

...

Supported Nextstrain environments: native, aws-batch

Setting default environment to docker.

WARNING: Default environment (docker) is not supported.
```

victorlin (changes requested):

nextstrain/cli/command/check_setup.py
tsibley: Yeah, I'll adjust the wording to that or something similar.

I'm also happy to replace "No good" with something else, but think `WARNING:` isn't quite the right replacement. But `ERROR:` or `Checks failed.` would be ok I think.

tsibley:
Hmm. With some other work on `nextstrain update` and a new `nextstrain setup`, I'm now reconsidering this implementation for `nextstrain check-setup`. Pondering implementing similar behaviour but in a different way that will make it parallel `update` and `setup` better.

PR #217, opened 2022-08-25
Standalone archive installers

tsibley:
Shell programs to download and unpack the standalone installation archives into standard locations, potentially upgrading/overwriting a prior standalone install.

These programs will be served from GitHub directly out of this repo via convenience redirects on nextstrain.org.

Related-to: <https://github.com/nextstrain/nextstrain.org/pull/582>
Related-to: <https://github.com/nextstrain/docs.nextstrain.org/pull/123>

### Todo

- [x] [Do more to adjust `PATH` on Linux and macOS?](https://github.com/nextstrain/cli/pull/217#issuecomment-1230851374)
- [x] Write tests for these in CI?

### Tests

- [x] Manually tested installer on Linux (@tsibley, you…?)
- [x] Manually tested installer on macOS (you…?)
- [x] Manually tested installer on Windows 10 (@tsibley, you…?)

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -1,0 +1,178 @@
+#!/bin/bash
+#
+# Bash program to download the latest Nextstrain CLI standalone installation
+# archive for Linux or macOS, extract it into the current user's app data
+# directory, and check if PATH includes the installation destination.
+#
+# It maintains rough parity with the PowerShell program for Windows,
+# standalone-installer-windows.
+#
+# Set DESTINATION to change the installation location.
+#
+# Set VERSION to change the version downloaded and installed, or pass the
+# desired version as the first argument to this program.
+#
+set -euo pipefail
+shopt -s failglob
+
+: "${NEXTSTRAIN_DOT_ORG:=https://nextstrain.org}"
+
+# Globals declared here to make them more obvious, but set by main() to avoid
+# source ordering requirements and delay execution until the full file is
+# parsed.
+declare KERNEL TARGET DESTINATION VERSION
+
+# Wrap everything in a function which we call at the end to avoid execution of
+# a partially-downloaded program.
+main() {
+    KERNEL="${KERNEL:-$(uname -s)}"
+    TARGET="${TARGET:-$(target-triple)}"
+    DESTINATION="${DESTINATION:-$(destination)}"
+
+    VERSION="${1:-${VERSION:-latest}}"
+
+    local archive archive_url tmp
+
+    archive="standalone-${TARGET}.tar.gz"
+    archive_url="${NEXTSTRAIN_DOT_ORG}/cli/download/${VERSION}/${archive}"
+
+    # Move into a temporary working dir
+    tmp="$(mktemp -d)"
+    if ! debug; then
+        trap "rm -rf $tmp" EXIT
+    fi
+    pushd "$tmp" >/dev/null
+    log "Temporary working directory: $tmp"
+
+    log "Downloading $archive_url"
+    curl "$archive_url" \
+        --fail --show-error --location --proto "=https" \
+        > "$archive"
+
+    log "Extracting $archive"
+    mkdir standalone
+    tar xz$(if-debug v)f "$archive" -C standalone
+
+    if [[ -d "$DESTINATION" ]]; then
+        log "Removing existing $DESTINATION"
+        rm -rf "$DESTINATION"
+    fi
+
+    log "Installing to $DESTINATION"
+    mkdir -p "$(dirname "$DESTINATION")"
+    mv standalone "$DESTINATION"
+
+    popd >/dev/null
+
+    if ! debug; then
+        log "Cleaning up"
+        rm -rf "$tmp"
+    fi
+
+    cat <<~~
+______________________________________________________________________________
+
+Nextstrain CLI ($("$DESTINATION"/nextstrain --version)) installed to $DESTINATION.
+
+~~
+
+    if [[ "$(type -p nextstrain 2>/dev/null)" != "$DESTINATION/nextstrain" ]]; then
+        cat <<~~
+To make the "nextstrain" command available without using its full path,
+please add the following to your shell initialization file (e.g. ~/.bashrc or
+~/.zshrc):
+
+    export PATH="$DESTINATION:\$PATH"
+
+and then logout and login again.  You only need to do this once.
+
+Your default shell appears to be $(default-shell).
+~~
+    fi
+}
+
+target-triple() {
+    local machine vendor os
+
+    machine="$(uname -m)"
+
+    case "$KERNEL" in
+        Linux)
+            [[ "$machine" == x86_64 ]] || die "unsupported architecture: $machine"
+            vendor=unknown
+            os=linux-gnu
+            ;;
+
+        Darwin)
+            [[ "$machine" == x86_64 || "$machine" == arm64 ]] || die "unsupported architecture: $machine"
+            machine=x86_64
+            vendor=apple
+            os=darwin
+            ;;
+
+        *)
+            die "unknown kernel: $KERNEL"
+    esac
+
+    echo "$machine-$vendor-$os"
+}
+
+destination() {
+    local user_data dest
+
+    # Paths from appdirs <https://github.com/ActiveState/appdirs/blob/master/appdirs.py>
+    case "$KERNEL" in
+        Linux)
+            user_data="${XDG_DATA_HOME:-$HOME/.local/share}";;
+
+        Darwin)
+            user_data="$HOME/Library/Application Support";;
+
+        *)
+            die "unknown kernel: $KERNEL"
+    esac
+
+    if [[ ! -d "$user_data" ]]; then
+        log "Current user's app data dir ($user_data) does not exist. Will create it." >&2
+    fi
+
+    echo "$user_data/nextstrain/cli/standalone"
+}
+
+default-shell() {
+    local shell
+
+    case "$KERNEL" in
+        Linux)
+            shell=$(getent passwd "$(id -u)" | awk -F: '{print $NF}');;
+
+        Darwin)
+            shell=$(dscl . -read ~/ UserShell | sed 's/UserShell: //');;
+
+        *)
+            die "unknown kernel: $KERNEL"
+    esac
+
+    basename "$shell"
+}
+
+debug() {
+    [[ -n "${DEBUG:-}" ]]
+}
+
+if-debug() {
+    if debug; then
+        echo "$@"
+    fi
+}
+
+log() {
+    echo "--> $@"
+}
+
+die() {
+    echo "ERROR:" "$@" >&2
+    exit 1
+}
+
+main "$@"
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -77,16 +77,20 @@ Nextstrain CLI ($("$DESTINATION"/nextstrain --version)) installed to $DESTINATIO
 ~~
 
     if [[ "$(type -p nextstrain 2>/dev/null)" != "$DESTINATION/nextstrain" ]]; then
+        local shell rc
+        shell="$(default-shell)"
+        rc="$(shell-rc "$shell")"
+
         cat <<~~
-To make the "nextstrain" command available without using its full path,
-please add the following to your shell initialization file (e.g. ~/.bashrc or
-~/.zshrc):
+To make the "nextstrain" command available in your default shell ($shell)
+without using the full path, please run these two commands now:
 
-    export PATH="$DESTINATION:\$PATH"
+    echo 'source <("$DESTINATION/nextstrain" init-shell $shell)' >> $rc
+    source <("$DESTINATION/nextstrain" init-shell $shell)
 
-and then logout and login again.  You only need to do this once.
-
-Your default shell appears to be $(default-shell).
+The first adds a line to your shell initialization file ($rc) for future
+sessions.  The second sets up your current shell session.  You only need to run
+these once.
 ~~
     fi
 }
@@ -153,7 +157,62 @@ default-shell() {
             die "unknown kernel: $KERNEL"
     esac
 
-    basename "$shell"
+    shell="$(basename "$shell")"
+
+    # Remove any -x.y version from the name
+    echo "${shell%%-*}"
+}
+
+shell-rc() {
+    local shell="$1"
+
+    # Paths below are presumed to be meta-char safe and returned with
+    # unexpanded ~/… prefixes for a nicer appearance in suggested commands,
+    # i.e. expansion of ~ is expected to happen after the user pastes and runs
+    # the commands, rather than here and now.
+    case "$shell" in
+        bash)
+            case "$KERNEL" in
+                Darwin)
+                    # macOS Terminal.app (and iTerm2.app) always launches login
+                    # shells, i.e. for all windows and tabs, so use the login
+                    # initialization file instead of the interactive one
+                    # because Bash (unlike zsh) only reads one or the other.
+                    # Although it's common practice for users to source their
+                    # bashrc from their bash_profile, we can't rely on this.
+                    #
+                    # The order and conditions on files here follows bash(1) §
+                    # INVOCATION:
+                    #
+                    #   After reading [/etc/profile], [bash] looks for
+                    #   ~/.bash_profile, ~/.bash_login, and ~/.profile, in that
+                    #   order, and reads and executes commands from the first one
+                    #   that exists and is readable.
+                    #
+                    if [[ -r ~/.bash_profile ]]; then
+                        echo "~/.bash_profile"
+
+                    elif [[ -r ~/.bash_login ]]; then
+                        echo "~/.bash_login"
+
+                    elif [[ -r ~/.profile ]]; then
+                        echo "~/.profile"
+
+                    else
+                        # If none exist, then we fallback to ~/.bash_profile
+                        # and expect it to be created by our suggested shell
+                        # init commands.
+                        echo "~/.bash_profile"
+                    fi;;
+                *)
+                    echo "~/.bashrc";;
+            esac;;
+        zsh)
+            echo "${ZDOTDIR:-~}/.zshrc";;
+        *)
+            # A decent guess?
+            echo "~/.${shell}rc";;
+    esac
 }
 
 debug() {
--- a/bin/standalone-installer-windows
+++ b/bin/standalone-installer-windows
@@ -1,0 +1,132 @@
+#
+# PowerShell 5+ program to download the latest Nextstrain CLI standalone
+# installation archive for Windows, extract it into the current user's app
+# data directory, and ensure PATH includes the installation destination.
+#
+# It maintains rough parity with the Bash program for Linux and macOS,
+# standalone-installer-unix.
+#
+# Set $env:DESTINATION to change the installation location.
+#
+# Set $env:VERSION to change the version downloaded and installed, or pass the
+# desired version as the first argument to this program.
+#
+Set-StrictMode -Version 3.0
+$ErrorActionPreference = "Stop"
+
+# Wrap everything in a function which we call at the end to avoid execution of
+# a partially-downloaded program.
+function main([string]$version) {
+    $destination = & {
+        if ($env:DESTINATION) {
+            return $env:DESTINATION
+        }
+        if (!$env:LOCALAPPDATA) {
+            throw "LOCALAPPDATA is not set"
+        }
+        if (!(Test-Path $env:LOCALAPPDATA)) {
+            throw "user's LOCALAPPDATA dir does not exist: ${env:LOCALAPPDATA}"
+        }
+        return "${env:LOCALAPPDATA}\nextstrain\cli\standalone"
+    }
+
+    $nextstrain_dot_org = & {
+        if ($env:NEXTSTRAIN_DOT_ORG) { return $env:NEXTSTRAIN_DOT_ORG }
+        return "https://nextstrain.org"
+    }
+
+    if (!$version) {
+        $version = & {
+            if ($env:VERSION) { return $env:VERSION }
+            return "latest"
+        }
+    }
+
+    # XXX TODO: Check for x86_64 arch; i.e. don't pass on 32-bit systems or
+    # non-Intel 64-bit (e.g. arm64).
+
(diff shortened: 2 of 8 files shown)


tsibley:
Open question: should we do more to update `PATH` for the user on Linux and macOS? Here's an example of what's output when `standalone-installer-unix` detects after installation that `nextstrain` isn't available (or isn't the one just installed):

```
Nextstrain CLI (nextstrain.cli 4.2.0) installed to /home/tom/.local/share/nextstrain/cli/standalone.

To make the "nextstrain" command available without using its full path,
please add the following to your shell initialization file (e.g. ~/.bashrc or
~/.zshrc):

    export PATH="/home/tom/.local/share/nextstrain/cli/standalone:$PATH"

and then logout and login again.  You only need to do this once.

Your default shell appears to be bash.
```

Some options:

1. Leave it as is.
2. Add additional text describing what to do in more detail.
3. Try to automatically modify the appropriate init file.

Option 3 can be done with various levels of sophistication, including:

- Prompting the user to proceed or not
- Distinguishing between default vs. current shell, particularly if they're not the same (e.g. we're running in bash but default shell is zsh)
- Installing/modifying init files for all detected _installed_ shells
- Future-proofing by recommending something like `source /home/tom/.local/share/nextstrain/cli/standalone/bashrc` or `source <(/home/tom/.local/share/nextstrain/cli/standalone/nextstrain init-shell bash)` instead of direct `PATH` modification.

I'd greatly appreciate thoughts, suggestions, and insight here on what may be best.

corneliusroemer:
Doesn't work for mac? Or am I doing something wrong?
image

corneliusroemer:
Ok, one issue could be that you give me a URL for arm - which doesn't exist in the releases. So either we need to add a convenience redirect for arm -> x86 or compile for arm.

image

tsibley:
@corneliusroemer Ah, yeah, was just separately realizing this as well as I thought through some other things. I don't have a Mac on which to test, so this "obvious" thing got overlooked.

corneliusroemer:
Ah good that I tested correctly then :) 

Since we only have x86 builds we can hardcode the `$machine` to that?

tsibley:
<https://duckduckgo.com/?q=uname+-m+on+apple+silicon> suggests DTRT may not be trivial... but I think since we're not building Nextstrain CLI for Apple Silicon yet, we could hardcode the architecture for Macs.

bin/standalone-installer-unix
```diff
@@ -0,0 +1,175 @@
...
+
+Your default shell appears to be $(default-shell).
+~~
+    fi
+}
+
+target-triple() {
+    local machine vendor os
+
+    machine="$(uname -m)"
```
corneliusroemer, line 97:
```suggestion
    machine="x86_64"
```

bin/standalone-installer-unix
```diff
@@ -0,0 +1,175 @@
...
+destination() {
+    local user_data dest
+
+    # Paths from appdirs <https://github.com/ActiveState/appdirs/blob/master/appdirs.py>
+    case "$KERNEL" in
+        Linux)
+            user_data="${XDG_DATA_HOME:-$HOME/.local/share}";;
+
+        Darwin)
+            user_data="$HOME/Library/Application Support";;
```
corneliusroemer, line 126:
```suggestion
            user_data="$HOME/Library/Application\ Support";;
```

Need to escape whitespace, otherwise this won't work with copy-paste

corneliusroemer (changes requested):
I find this blurb in the end hard to parse - it would be good to insert paragraph breaks

See screenshot:
image

bin/standalone-installer-unix
tsibley: I'll modify to use `x86_64` unconditionally on `darwin` systems, but check `uname -m` and error if it's not `x86_64` on others.

bin/standalone-installer-unix
tsibley: I don't follow. Copying what from where?

tsibley:
> I find this blurb in the end hard to parse - it would be good to insert paragraph breaks

Hmm. This must be a zsh thing? You can see how it looks for me with bash (on Linux) in [this earlier comment](https://github.com/nextstrain/cli/pull/217#issuecomment-1230851374).

bin/standalone-installer-unix
tsibley: [Done](https://github.com/nextstrain/cli/compare/bcf777a286854b2f5de4827fb7d984ebb2b2204d..e32a20cba15114ccbe8009fd9276079ab1152051).

tsibley:
Doesn't appear to be zsh, at least when I test with zsh locally. So I'm confused about the missing blank lines, as they're definitely there in the source of the heredoc.

tsibley:
(and of course, the code is running as bash, not zsh… I keyed on your default shell being different, but it doesn't matter. so maybe your terminal is doing something weird?)

bin/standalone-installer-unix
corneliusroemer: Sorry my bad. My suggestion doesn't actually work.

The problem is that the blurb on how to add the command to PATH didn't work when I copied without `"`.

But that was my mistake. Both of the following work:
```
/Users/corneliusromer/Library/Application\ Support/nextstrain/cli/standalone/nextstrain   
"/Users/corneliusromer/Library/Application Support/nextstrain/cli/standalone/nextstrain"
```

But the following do not work:
```
/Users/corneliusromer/Library/Application Support/nextstrain/cli/standalone/nextstrain   
"/Users/corneliusromer/Library/Application\ Support/nextstrain/cli/standalone/nextstrain"
```

corneliusroemer: Talking about this part image

tsibley: Ah, got it. No worries.

bin/standalone-installer-unix
```diff
@@ -0,0 +1,178 @@
...
+
+        *)
+            die "unknown kernel: $KERNEL"
+    esac
+
+    if [[ ! -d "$user_data" ]]; then
+        die "user's app data dir does not exist: $user_data"
+    fi
```
tsibley, line 137: I had an `XXX FIXME should we create it?` comment to myself here before opening this PR that I answered with "nah" and removed. But now the testing on Ubuntu 18.04 and 20.04 show that maybe the answer should be "jah": https://github.com/nextstrain/cli/runs/8081754539?check_suite_focus=true#step:3:5

tsibley: Fixed up in 4432b0c.

bin/standalone-installer-unix
```diff
@@ -0,0 +1,178 @@
...
+
+    if [[ "$(type -p nextstrain 2>/dev/null)" != "$DESTINATION/nextstrain" ]]; then
+        cat <<~~
+To make the "nextstrain" command available without using its full path,
+please add the following to your shell initialization file (e.g. ~/.bashrc or
+~/.zshrc):
+
+    export PATH="$DESTINATION:\$PATH"
+
+and then logout and login again.  You only need to do this once.
```
victorlin, line 87: Re: [should we do more to update `PATH`?](https://github.com/nextstrain/cli/pull/217#issuecomment-1230851374)

Two things that I've used personally come to mind:

1. Homebrew outputs a command at the end of the installation script that will append the necessary init command to the shell initialization file ([src](https://github.com/Homebrew/install/blob/805136bcfcbc40b95e476639fb2407773ab26445/install.sh#L995-L1021)).
2. Conda has its own `conda init` command which then writes a conditional block to the shell initialization file ([src](https://github.com/conda/conda/blob/c031405723cbf934de1a0b0c645b21e4c6fe5a7f/conda/core/initialize.py#L433)).

I think it'd be good to simplify down to one script block that the user can copy/paste and run without replacing any text (similar to Homebrew's behavior, and kinda option 2 from the linked comment). It would:

1. Detect the shell(s)
2. Determine the correct shell initialization file(s)
3. Append `PATH` modification to shell initialization file(s)

tsibley: Thanks for your thoughts here! This seems reasonable to me.

So we're aiming for something like [option 2 in my comment](https://github.com/nextstrain/cli/pull/217#issuecomment-1230851374). Should we take on any of the "optional sophistications" I noted which could apply here?

- Distinguishing between default vs. current shell, particularly if they're not the same (e.g. we're running in bash but default shell is zsh)
- Instructions for installing/modifying init files for all detected _installed_ shells (i.e. not just the default or current one)
- Future-proofing by recommending something like `source /home/tom/.local/share/nextstrain/cli/standalone/bashrc` or `source <(/home/tom/.local/share/nextstrain/cli/standalone/nextstrain init-shell bash)` instead of direct `PATH` modification.

I'm inclined to do the future-proofing bit to simplify (hypothetical) future changes for Linux/macOS users as they wouldn't have to update their shell init again, but the downside is it complicates our Windows install, as currently we can just automatically make the PATH adjustments in a shell-agnostic way. I'm also suspicious that the future-proofing would be worth it, as I can't currently imagine what other bits of shell init/config we'd need to do besides `PATH` (but my imagination may be limited).

Thoughts on any of the above? In particular, I wonder about the default vs. current shell question.

victorlin: > default vs. current shell question

Looks like this is what conda does or is capable of doing ([src](https://github.com/conda/conda/blob/c031405723cbf934de1a0b0c645b21e4c6fe5a7f/conda/core/initialize.py#L436)). I think it's a good idea if not too difficult.

> Future-proofing by recommending something like `source /home/tom/.local/share/nextstrain/cli/standalone/bashrc`

Do you mean having the user run `source /home/tom/.local/share/nextstrain/cli/standalone/bashrc` which then modifies `.bashrc`, or putting the one line in `.bashrc` itself?

I think the one line in `.bashrc` itself would be good. To add another example, `rbenv` added just one line to my `.zshrc` which I think is nice ([src](https://github.com/rbenv/rbenv/blob/585ed84283f3308380b843391ee7b12706ecff87/libexec/rbenv-init#L77)):

```sh
eval "$(rbenv init - zsh)"
```

tsibley: Ah, so the functionality of the conda source you linked is because you can do `conda init bash zsh` for example and it'll go and modify the config for each named shell. You can also ask it to modify the config for all shells it supports (quite a few). I'm asking about something a bit different. We need to detect the shell (X) to emit instructions for when we print something like, "Add this line to your X shell config: {instructions for X}". Should we detect the _current_ shell or the _default_ shell? This is a subtle thing, but they're not the same and we need to make a choice. I'd lean towards default shell.

Re: future-proofing: I don't mean having the user source something to do the modification for them. I mean the difference between instructing the user to add a line like this to their shell config:

    export PATH="/home/tom/.local/share/nextstrain/cli/standalone:$PATH"

or a line like this instead:

    source /home/tom/.local/share/nextstrain/cli/standalone/etc/bashrc

or maybe like this:

    source <(/home/tom/.local/share/nextstrain/cli/standalone/nextstrain init-shell bash)

The potential future-proofing benefit of the last two is that the necessary shell init is automatically updated when Nextstrain CLI is upgraded. With the first, if changes are ever necessary, we'd need to instruct users again to manually make them.

tsibley: Answering one of my own questions:

```
source <(/home/tom/.local/share/nextstrain/cli/standalone/nextstrain init-shell bash)
```

is preferable because we can write more complex conditional/detection logic in Python and emit simpler POSIX shell syntax instead of having to write separate shell scripts for each shell flavor.

tsibley: > Should we detect the _current_ shell or the _default_ shell? This is a subtle thing, but they're not the same and we need to make a choice. I'd lean towards default shell.

Oh right, nevermind. The _current_ shell in this installer is always going to be Bash because that's what we run the installer under. So it makes no sense to consider it! ISTR realizing this a week or two ago but clearly I'd forgotten again.

tsibley: Pushed a81d314 addressing this.

nextstrain/cli/command/init_shell.py
```diff
@@ -0,0 +1,102 @@
...
+except:
+    INSTALLATION_PATH = None
+
+__doc__ = (__doc__ or "").format(
+    INSTALLATION_PATH = (
+             shquote(str(INSTALLATION_PATH))
+          if INSTALLATION_PATH
+        else "…/path/to"
+    )
+)
```
victorlin, line 35: Curious, why `__doc__ or ""` instead of just `__doc__`?

victorlin (approved):
Tested on macOS (M1, zsh) with Docker runtime:

- [x] Installation works (see inline comment below)
- [x] `nextstrain check-setup --set-default` works
- [x] `nextstrain build --cpus 1 zika-tutorial/` works
- [x] `nextstrain view zika-tutorial/auspice/` works

bin/standalone-installer-unix
victorlin: I see, default shell is good then! Seems like it'd be easy to apply to all installed shells if wanted in the future.

nextstrain/cli/command/init_shell.py
tsibley: It's a minor wart solely to satisfy type checking since pyright thinks `__doc__` might be `None`.

```shell-session
$ cat /tmp/doctype.py 
"""
Hello, docstring!
"""
reveal_type(__doc__)

$ mypy /tmp/doctype.py 
/tmp/doctype.py:4: note: Revealed type is "builtins.str"
Success: no issues found in 1 source file

$ pyright /tmp/doctype.py 
Loading configuration file at /home/tom/nextstrain/cli/pyrightconfig.json
Assuming Python platform Linux
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding **/.*
stubPath /home/tom/nextstrain/cli/typings is not a valid directory.
Searching for source files
Found 1 source file
pyright 1.1.270
/tmp/doctype.py
  /tmp/doctype.py:4:13 - information: Type of "__doc__" is "str | None"
0 errors, 0 warnings, 1 information 
Completed in 0.599sec
```

nextstrain/cli/__init__.py
```diff
@@ -76,6 +76,7 @@ def make_parser():
         logout,
         whoami,
         version,
+        init_shell,
```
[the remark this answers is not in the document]
tsibley: :raised_hands: I should have mentioned that I tested by building the standalone installation locally with `pyoxidizer build` and then invoking `./build/x86_64-unknown-linux-gnu/debug/installation/nextstrain init-shell`.

bin/standalone-installer-windows
```diff
@@ -0,0 +1,132 @@
+#
```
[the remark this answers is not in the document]
tsibley: I think it belies that you could run it on Unix under pwsh, when in fact you can't. Plus, the very next line already says it's a PowerShell program:

https://github.com/nextstrain/cli/blob/40724d70fee962f43bfa5c3a02eff1f134701306/bin/standalone-installer-windows#L1-L2

nextstrain/cli/command/init_shell.py
victorlin: Ohh I see. I guess it'd be clear when CI fails, but a comment wouldn't hurt. This sounds like a bad discrepancy between mypy and pyright. Not sure which one is actually correct, they're both hardcoded...

```
% cat no_docstring.py 
reveal_type(__doc__)

% mypy no_docstring.py 
no_docstring.py:1: note: Revealed type is "builtins.str"

% npx pyright no_docstring.py
...
  /tmp/no_docstring.py:1:13 - information: Type of "__doc__" is "str | None"
0 errors, 0 warnings, 1 information 
```

tsibley: I'll add a comment.

Some context on the pyright and mypy behaviour here:

- https://github.com/microsoft/pyright/issues/2241
- https://github.com/python/mypy/issues/9170
- https://github.com/python/mypy/issues/2709

from which it sounds like recent mypy's should be typing `__doc__` equivalently as `Optional[str]`, so not sure why we're not observing that (using most recent version on PyPI).

tsibley: Ok, the different is the `strict_optional = False` in our `mypy.ini`. Running mypy outside of the repo with `--no-strict-optional` produces the same result.

tsibley: https://github.com/python/mypy/blob/7c14feedd2a6889d9eab8b0ac8dc8aab630bbed3/mypy/nodes.py#L119-L126

https://github.com/python/mypy/blob/7c14feedd2a6889d9eab8b0ac8dc8aab630bbed3/mypy/semanal.py#L595-L599

tsibley: The Python data model [is pretty clear](https://docs.python.org/3/reference/datamodel.html?highlight=__doc__#the-standard-type-hierarchy) that a module's `__doc__` may be `None`.

image

joverlee521:
Tested with bash on macOS v10.14.6 and everything worked for my current shell session.
However, my terminal creates every new tab/window as a login shell so `nextstrain init-shell` was not invoked for a new session. Not sure how much hand-holding we want to do here for the older macOS versions.

tsibley:
> However, my terminal creates every new tab/window as a login shell so `nextstrain init-shell` was not invoked for a new session. Not sure how much hand-holding we want to do here for the older macOS versions.

Oh, hmm. I'd think the older macOS versions aren't that old and thus will be prevalent and thus we should take them into account. But I don't have the best sense of this here being an outsider in that ecosystem.

So the default system profile (e.g. `/etc/profile` by default on bash, but may be set different by Apple) doesn't source the rc? That's a pretty common system-level default config, and I'm a little surprised it's not the case here.

Is your shell bash or zsh?

joverlee521:
> So the default system profile (e.g. `/etc/profile` by default on bash, but may be set different by Apple) doesn't source the rc? That's a pretty common system-level default config, and I'm a little surprised it's not the case here.

Maybe this is just weird behavior with the Terminal app? There's [common advice](https://apple.stackexchange.com/a/271472) to source `~/.bashrc` within `~/.bash_profile`,  but I'm not sure about the default system profile. 

> Is your shell bash or zsh?

I'm using bash

tsibley:
Ok, so your `~/.bash_profile` (or `~/.bash_login`, or `~/.profile`) doesn't source `~/.bashrc` then?

joverlee521:
> Ok, so your `~/.bash_profile` (or `~/.bash_login`, or `~/.profile`) doesn't source `~/.bashrc` then?

It does not. I didn't even have a `~/.bashrc` file prior to running the recommend steps from the installer 😆

victorlin:
I vaguely remember this behavior from using earlier versions of macOS. Lots of StackExchange posts related to this [[1]] [[2]]. `.zshrc` gets sourced by default with newer versions of macOS though.

It's also interesting that [this Apple help page](https://support.apple.com/en-us/HT208050) says

> **.zshrc** is equivalent to **.bashrc** and runs for each new Terminal session

when clearly, .bashrc didn't run for each new Terminal session unless you explicitly told it to.

[1]: https://apple.stackexchange.com/q/119711

[2]: https://apple.stackexchange.com/q/12993

tsibley:
Ah, ok. I first thought maybe Terminal.app was starting login shells on older macOS versions and not on newer macOS versions, but it turns out there's a fundamental difference in startup behaviour between zsh and bash: bash only executes the system and user rc files "[w]hen an interactive shell that is not a login shell is started". zsh executes the system and user rc files for any interactive shell, including login shells.

So, thinking aloud… for bash users, should we check or modify the profile/login startup files for sourcing of bashrc? Or advise to check that in an installation troubleshooting section? Or be doing something entirely different here?

tsibley:
Ahh, and now it clicks why Conda uses `~/.bash_profile` on macOS (and Windows) but `~/.bashrc` elsewhere ([ref](https://github.com/conda/conda/blob/c031405723cbf934de1a0b0c645b21e4c6fe5a7f/conda/core/initialize.py#L438-L439)).

tsibley:
@victorlin @joverlee521 Do alternate terminal emulators like iTerm2.app also start login shells by default? If so, then probably safe to assume we should use the profile/login startup file for bash on macOS instead.

tsibley:
Answering my own question [from documentation](https://iterm2.com/documentation-preferences-profiles-general.html), it appears so:

> **Command**
>
> This is the command that is executed when a new session with the profile is created. If _login shell_ is chosen, then `login` is invoked. You can put special terms surrounded by `$$` in this field (example: `$$USERNAME$$`). When a new session is created, you will be prompted to enter a value for each such term. See the description of URL Schemes below for details about the special `$$` value that can go in this field.
> 
> When _custom shell_ is selected, you should enter the path to a shell (e.g., `/usr/local/bin/bash`) and it will be run as a login shell.

bin/standalone-installer-windows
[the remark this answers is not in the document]
tsibley: If this is more about editor config than human's reading, I did put a vim modeline at the bottom:

https://github.com/nextstrain/cli/blob/40724d70fee962f43bfa5c3a02eff1f134701306/bin/standalone-installer-windows#L132

but this will only work with compatible modeline-supporting editors. We could switch to a shebang if that's better, I suppose.

nextstrain/cli/command/init_shell.py
tsibley: Comment added in f6e2c8a.

tsibley:
@joverlee521 Try the new revision of the installer?

tsibley:
One thing I realized the other day and am thinking about now is that on Linux/macOS we execute the installer in isolation, but on Windows we effectively `source` it. This lets the Windows installer modify the current shell session directly (separately from modifying the persistent `PATH` value). If we switched to sourcing the Unix installer too, then we could eliminate the second command here:

https://github.com/nextstrain/cli/blob/35fe49ccdf27cf1e63ad43970b063f35c6db5c4b/bin/standalone-installer-unix#L85-L93

…but we can't reliably source it because we don't know what the shell will be, whereas on Windows we mandate/assume PowerShell for the install (though once installed, `nextstrain` is usable in cmd.exe too).

Anyhow, I wanted to jot down these thoughts somewhere while they were back in my head. I don't think anything needs to change at the moment.

nextstrain/cli/command/init_shell.py
victorlin: Haha, today's Python learning! Thanks for digging and adding the comment.

PR #222, opened 2022-10-06
runner.docker: Initial, minimal `setup` support

tsibley:
Downloads the image if it's not already available locally, avoiding an awkward download on first use.  It makes more sense to do this in `setup` than to advise folks to run `update` after install.  The update machinery is still used behind the scenes.

Diff best viewed with whitespace ignored.

### Testing
- [x] CI passes, though why wouldn't it?
- [x] Locally tested by hand

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,13 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Improvements
+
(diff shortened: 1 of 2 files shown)


victorlin (approved):
- [x] `nextstrain setup docker` looks good
- [x]  `nextstrain setup docker --force` looks good
- [x]  `nextstrain setup docker --dry-run` looks good

PR #224, opened 2022-10-06
Rename "native" runtime/runner to "ambient"

tsibley:

### Related issue(s)
Related to #218.

### Testing
- [x] Local tests pass
- [x] Manually driving commands works ok
- [x] CI passes

### DIFF
--- a/nextstrain/cli/argparse.py
+++ b/nextstrain/cli/argparse.py
@@ -175,12 +175,12 @@ def runner_module_argument(name: str) -> RunnerModule:
     >>> runner_module_argument("invalid")
     Traceback (most recent call last):
         ...
-    argparse.ArgumentTypeError: invalid runtime name: 'invalid'
+    argparse.ArgumentTypeError: invalid runtime name: 'invalid'; valid names are: 'docker', 'conda', 'native', 'aws-batch'
 
     >>> runner_module_argument("Invalid Name")
     Traceback (most recent call last):
         ...
-    argparse.ArgumentTypeError: invalid runtime name: 'Invalid Name' (normalized to 'invalid-name')
+    argparse.ArgumentTypeError: invalid runtime name: 'Invalid Name' (normalized to 'invalid-name'); valid names are: 'docker', 'conda', 'native', 'aws-batch'
     """
     try:
         return runner_module(name)
--- a/nextstrain/cli/argparse.py
+++ b/nextstrain/cli/argparse.py
@@ -1,15 +1,13 @@
 """
 Custom helpers for extending the behaviour of argparse standard library.
 """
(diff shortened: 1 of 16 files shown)


nextstrain/cli/argparse.py
```diff
@@ -157,43 +155,37 @@ def __call__(self, parser, namespace, value, option_string = None):
...
 
-    >>> runner_module("invalid")
+    >>> runner_module_argument("native") # doctest: +ELLIPSIS
+    <module 'cli.runner.ambient' from '...'>
+
+    >>> runner_module_argument("invalid")
     Traceback (most recent call last):
         ...
-    argparse.ArgumentTypeError: invalid runtime name: 'invalid'
+    argparse.ArgumentTypeError: invalid runtime name: 'invalid'; valid names are: 'docker', 'conda', 'ambient', 'aws-batch'
```
victorlin, line 181: 17841d4 includes `ambient` in parts of this file when it should be `native` first then renamed in c7ee3cc.

nextstrain/cli/util.py
```diff
@@ -339,6 +339,60 @@ def runner_name(runner: RunnerModule) -> str:
...
+    ValueError: invalid runtime name: 'Invalid Name' (normalized to 'invalid-name'); valid names are: 'docker', 'conda', 'ambient', 'aws-batch'
+    """
+    # Import here to avoid circular import
+    from .runner import all_runners_by_name
+
+    normalized_name = re.sub(r'(_|\s+)', '-', name).lower()
+
+    # Accept "native" as an alias for "ambient" for backwards compatibility
+    if normalized_name == "native":
+        normalized_name = "ambient"
```
victorlin, line 383: Could we add a DeprecationWarning here? Mostly to inform existing `native` users of the new name, but also to justify eventual removal of the alias.

tsibley: I'd like not to. I'm not a fan of nagging users about things that still work just fine, and I wouldn't plan to remove `--native`. It is very little overhead to keep it around as an alias, we're unlikely to want to re-use it for something else in the future, and keeping it around will avoid breaking users' stuff.

If we do want to warn, we wouldn't use a DeprecationWarning, because [they're intended for developers and aren't visible by default to users](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

nextstrain/cli/argparse.py
tsibley: Whoops! Good catch. There was another anachronism in 8f1aa16 too. I was moving too fast at the end of the day! [Fixed with repush](https://tsibley.net/paste/2022-10-11IAmVQedu).

Issue #231, opened 2022-10-31
[shell] shell-history does not exist for Docker


joverlee521:
### Current Behavior
When running `nextstrain shell` with Docker, the command errors if the shell-history file does not exist:
```
$ nextstrain shell .
Entering the Nextstrain build environment

Mapped volumes:
  /nextstrain/build is from /Users/jlee2346/Repos/nextstrain

Run the command "exit" to leave the build environment.

Error: The path(s) given for the following components do not exist
or are not directories:

    • bash_history: /Users/jlee2346/.nextstrain/shell-history
```

### Expected behavior
The shell-history file should be created prior to mounting in Docker. 
Should the cli automatically create the file behind the scenes or should we just leave it to users to create it?

tsibley:
Oh, bummer. This is a regression in 5.0.0 compared to 4.2.0. The error comes from `nextstrain.cli.runner.docker.run()`, which checks that all mounted volumes from the host already exist:

https://github.com/nextstrain/cli/blob/4e1960bb64b039d54fc9a4d52c4504c7004613cd/nextstrain/cli/runner/docker.py#L63-L79

tsibley:
I think `nextstrain.cli.command.shell.run()` should explicitly ensure that the `history_file` it arranges exists with something like:

```python
history_file = SHELL_HISTORY.resolve(strict = True)
history_file.parent.mkdir(parents = True, exist_ok = True)

try:
    history_file.touch()
except FileExistsError:
    pass
```

tsibley:
Should be fixed by #232.

tsibley:
Fix will be released as 5.0.1 when it gets thru CI.

PR #232, opened 2022-11-01
shell: Ensure the history file always exists

tsibley:
Resolves <https://github.com/nextstrain/cli/issues/231>.

### Testing

- [x] Checks pass
- [x] Works locally
- [x] Works for @joverlee521

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,13 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Bug fixes
+
+* `nextstrain shell` no longer errors when its history file, e.g.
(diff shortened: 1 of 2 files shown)


tsibley:
The easiest way to test this is by setting `NEXTSTRAIN_HOME` to a directory that doesn't exist and/or a directory that does exist, but is missing a `shell-history` file.

joverlee521 (approved):
Thanks for fixing this! 

> The easiest way to test this is by setting NEXTSTRAIN_HOME to a directory that doesn't exist and/or a directory that does exist, but is missing a shell-history file.

Good to know! I totally missed `NEXTSTRAIN_HOME` was added!

tsibley:
This fix will be released as 5.0.1, currently building.

PR #238, opened 2022-11-23
remote.nextstrain_dot_org: Handle 400 Bad Request errors

tsibley:
This helps in particular with user-caused bad requests, such as trying to upload to dataset URLs which include underscores instead of slashes. It will also let us use 400 responses more widely on the server side since we know the CLI user will now see the error message.

Related-to: https://github.com/nextstrain/nextstrain.org/pull/611

### Testing

Example of the error message:

```
$ NEXTSTRAIN_DOT_ORG=http://localhost:5000 nextstrain remote upload groups/test/foo_bar foo_bar
.json
Uploading foo_bar.json as http://localhost:5000/groups/test/foo_bar
Error: The remote server rejected our request:

  Resource (e.g. dataset and narrative) paths may not include
  underscores (_); use slashes (/) instead

This may indicate a problem that's fixable by you, or it
may be a bug somewhere that needs to be fixed by the
Nextstrain developers.

If you're unable to address the problem, please open a new
issue at <https://github.com/nextstrain/cli/issues/new/choose>
and include the complete output above and the command you
were running.
```

- [x] Checks pass
- [x] Manual testing of 400 with decodable JSON body
- [x] Manual testing of 400 with bad JSON body

PR #239, opened 2022-11-23
Fix CI (again)

tsibley:
Fixes syntax of the flake8 config file.

The trailing comments were apparently bogus this whole time!  The
configuration documentation does mention this¹, though not sure it did
at the time I wrote the config initially.  In any case: oops.

flake8 version 6.0.0, released earlier today, started validating the
config and thus detecting the erroneous entries, which made our CI start
to fail with:

    ValueError: Error code '#' supplied to 'extend-ignore' option does
    not match '^[A-Z]{1,3}[0-9]{0,3}$'

Failures were limited to Python ≥3.8, as flake8 6.0.0 bumped its minimum
Python dep, so older Python versions kept using older flake8 versions,
but the syntax error affected them nevertheless.

¹ https://flake8.pycqa.org/en/latest/user/configuration.html

### Related issue(s)

CI failures [noticed in an unrelated PR](https://github.com/nextstrain/cli/pull/238#issuecomment-1325665456).

### Testing

- [ ] Checks pass

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,11 +3,15 @@
 
 [flake8]
 select =
-    E9,     # syntax errors
-    F       # all pyflakes correctness issues
+    # syntax errors
+    E9,
+
+    # all pyflakes correctness issues
+    F,
 
 extend-ignore =
-    F541    # allow f-strings without any placeholders
+    # allow f-strings without any placeholders
+    F541,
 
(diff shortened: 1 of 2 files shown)


tsibley:
I'm going to merge this once CI passes for the PR.

PR #247, opened 2023-01-04
standalone-installer-unix: Error if Rosetta 2 is not enabled on arm64 Macs

tsibley:
Rosetta 2 is required for both Nextstrain CLI (which doesn't yet have arm64 binaries) and our runtimes.

While we expect most Nextstrain users on arm64 Macs will already have Rosetta 2 enabled, it's good to check and provide a way forward instead of erroring during our installation check about a "Bad CPU type in executable".  In particular, anyone who's setting up a new Mac and installing Nextstrain as one of their first activities is likely to encounter the issue.

If necessary, this check can be bypassed by manually specifying

    TARGET=x86_64-apple-darwin

when running the installer.  This provides an escape hatch we can immediately provide to users should our `pgrep` check turn bad due to changes in Rosetta or macOS.

Related-to: <https://github.com/nextstrain/docs.nextstrain.org/pull/147>

### Testing

- [ ] Works on an arm64 Mac with Rosetta enabled
- [ ] Works on an arm64 Mac with Rosetta disabled (maybe hard to test…)
- [ ] Checks pass

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -115,6 +115,7 @@ target-triple() {
 
         Darwin)
             [[ "$machine" == x86_64 || "$machine" == arm64 ]] || die "unsupported architecture: $machine"
+            [[ "$machine" != arm64 ]] || pgrep -qf /usr/libexec/rosetta/oahd || die "Rosetta 2 not enabled.  Please run:"$'\n\n'"    softwareupdate --install-rosetta"$'\n\n'"and then retry this installation of Nextstrain CLI."
             machine=x86_64
             vendor=apple
             os=darwin


tsibley:
Tested "Rosetta disabled" by killing its translation process:

    sudo pkill -f /usr/libexec/rosetta/oahd

and then running the installer:

    curl -fsSL --proto '=https' https://raw.githubusercontent.com/nextstrain/cli/7c88c63cff3cd7dc4d9d04094730c0d55ae5be38/bin/standalone-installer-unix | bash -s 6.0.2

victorlin:
> Tested "Rosetta disabled" by killing its translation process

I tried this on my M1 mac but it re-spawns, possibly upon usage?

```
% conda config --env --get subdir \
    && sudo pkill -f /usr/libexec/rosetta/oahd \
    && ps aux | grep rosetta \
    && python -c "import platform;print(platform.machine())" \
    && ps aux | grep rosetta
--set subdir osx-64
vlin              6990   0.0  0.0 407962448     48 s000  S+    1:26PM   0:00.00 grep rosetta
x86_64
vlin              6994   0.0  0.0 407962448     48 s000  S+    1:26PM   0:00.00 grep rosetta
_oahd             6992   0.0  0.0 408501280   2176   ??  Ss    1:26PM   0:00.01 /usr/libexec/rosetta/oahd
```

tsibley:
@victorlin When I tested on `blab-mac`, it didn't automatically respawn on me until I ran an x86_64 executable. I think the `python` invocation in your example is triggering the respawn.

tsibley:
To clarify, the installer detects if the Rosetta 2 translation process is _running_. If it's not running, it assumes it's not installed and requests installation first. Re-installing is harmless (except for some wasted time/bandwidth) if Rosetta 2 was in fact already installed but merely not running for some reason. So my test with `pkill` was specifically to test the detection logic of the installer.

victorlin:
Oh right, that makes sense. Thanks for clarifying!

victorlin (approved):

Issue #2, opened 2018-06-29
Support using Singularity to run builds in the container image


tsibley:
`nextstrain build` should be able to support running builds using the container image but in an environment which doesn't support docker or have it installed, such as the Hutch's rhino cluster.  This can be accomplished using [Singularity](https://singularity.lbl.gov).  The interface might look like one of these:

```
nextstrain build --singularity .
```

Implementing this would involve adding a new runner, nextstrain/cli/runner/singularity.py, and registering it within the runner framework.

pvanheus:
Hi there 

I'll look into adding it to the wrapper, but if you want to use Singularity with nextstrain you can:

1) build a singularity image (I used a specific tagged version for this, see [Docker Hub](https://hub.docker.com/r/nextstrain/base/tags) for list of tags:

```
singularity build nextstrain-base-build-20200309T035736Z.simg 'docker://nextstrain/base:build-20200309T035736Z'
```
2) running the `ncov` pipeline from within a cloned copy of the [ncov](https://github.com/nextstrain/ncov) github repository, this does `nextstrain build .`

```
singularity exec -B $(pwd):/nextstrain/build nextstrain-base-build-20200309T035736Z.simg  snakemake --printshellcmds
```
For reference, the Docker command that this `nextstrain build .` runs is:

```
docker run --rm --tty --interactive --user=$(id -u):$(id -g) --volume=$(pwd):/nextstrain/build --workdir=/nextstrain/build --env=AUGUR_RECURSION_LIMIT --env=AUGUR_MINIFY_JSON --env=AWS_ACCESS_KEY_ID --env=AWS_SECRET_ACCESS_KEY --env=AWS_SESSION_TOKEN --env=ID3C_URL --env=ID3C_USERNAME --env=ID3C_PASSWORD --env=RETHINK_HOST --env=RETHINK_AUTH_KEY nextstrain/base snakemake --printshellcmds
```

tsibley:
Thanks for looking into this @pvanheus! If you have any questions about how the CLI works, feel free to ask them here.

pvanheus:
This works fairly similarly for ncov-ingest:

```bash
SINGULARITYENV_GISAID_API_ENDPOINT=https://www.epicov.org/BLAH
SINGULARITYENV_GISAID_USERNAME_AND_PASSWORD=USER:PASSWORD
singularity exec /tools/containers/nextstrain/ncov-ingest.sif snakemake --configfile config/local_gisaid.yaml
```

[...]

PR #248, opened 2023-01-06
Singularity runtime

tsibley:
Add a new runner for a Singularity runtime based on the Docker runtime

It's been my intention to support Singularity since day 1, but I never got around to implementing it because usage seemed little to non-existent amongst Nextstrain users I knew about.  (Though maybe that's a chicken-or-the-egg problem!)  We've since had users report their ad-hoc use of Singularity with Nextstrain, however, and Singularity is often the only containerization method supported by HPC systems.  So it seems important to support.

Resolves <https://github.com/nextstrain/cli/issues/2>.

### Testing

You can test this yourself using a source checkout of this PR, or with the standalone build produced by this PR:

```
curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux \
    | DESTINATION=/tmp/cli-test bash -s pr-build/248
```

(Download will be much slower than a normal release, but that's ok for dev.)

No need to run the two final commands the installer suggests at the end.

Then get started testing it with:

```
/tmp/cli-test/nextstrain setup singularity
```

### Checks

- [x] `check-setup` works
- [x] `setup` works
- [x] `update` works
- [x] `build` works
- [x] `view` works
- [x] `shell` works
- [x] Tests pass locally
- [x] Checks pass

### DIFF
--- a/.github/actions/run-integration-tests/action.yaml
+++ b/.github/actions/run-integration-tests/action.yaml
@@ -23,6 +23,10 @@ runs:
       shell: bash -l -eo pipefail {0}
       run: nextstrain setup conda
 
+    - if: runner.os != 'macOS' && runner.os != 'Windows'
+      shell: bash -l -eo pipefail {0}
+      run: nextstrain setup singularity
+
     - shell: bash -l -eo pipefail {0}
       run: nextstrain check-setup --set-default
 
@@ -42,6 +46,13 @@ runs:
       run: |
         git -C zika-tutorial clean -dfqx
         nextstrain build --conda --cpus 2 zika-tutorial
+
+    - if: runner.os != 'macOS' && runner.os != 'Windows'
+      name: Build zika-tutorial with --singularity
+      shell: bash -l -eo pipefail {0}
+      run: |
+        git -C zika-tutorial clean -dfqx
+        nextstrain build --singularity --cpus 2 zika-tutorial
 
     - if: runner.os != 'Windows'
       name: Build zika-tutorial with --ambient
--- a/.github/actions/setup-integration-tests/action.yaml
+++ b/.github/actions/setup-integration-tests/action.yaml
@@ -68,6 +68,34 @@ runs:
         python3 --version | grep -F 'Python ${{ inputs.python-version }}.'
         [[ "$(python --version)" == "$(python3 --version)" ]]
 
+    # Install Singularity on Linux.
+    #
+    # We don't install it with Conda because Conda Forge provides a non-suid
+    # build of Singularity.  We're compatible with Singularity's non-suid mode,
+    # but production usages of Singularity are likely to use its suid mode, so
+    # I'd rather test against that.
+    #   -trs, 6 Jan 2023
+    - if: runner.os == 'Linux'
+      shell: bash -l -eo pipefail {0}
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        # Work in a temp dir to avoid cluttering the caller's working dir
+        pushd "$(mktemp -d)"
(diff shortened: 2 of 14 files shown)


j23414:
Nice to see, especially for HPC users where [Docker is not allowed for security reasons](https://researchcomputing.princeton.edu/support/knowledge-base/singularity#docker).

Although I'll be curious how you handle [sysadmin defined bind paths](https://docs.sylabs.io/guides/3.0/user-guide/bind_paths_and_mounts.html#system-defined-bind-paths).

tsibley:
> Although I'll be curious how you handle [sysadmin defined bind paths](https://docs.sylabs.io/guides/3.0/user-guide/bind_paths_and_mounts.html#system-defined-bind-paths).

The runner uses Singularity's `--contain` option to avoid most system-defined bind paths. Per the default `singularity.conf`'s commentary on the system `bind path` config:

    # NOTE: these are ignored if singularity is invoked with --contain except
    # for /etc/hosts and /etc/localtime. When invoked with --contain and --net,
    # /etc/hosts would contain a default generated content for localhost resolution.

So I think system-defined bind paths should be a non-issue!

tsibley:
> Nice to see, especially for HPC users where [Docker is not allowed for security reasons](https://researchcomputing.princeton.edu/support/knowledge-base/singularity#docker).

Yep! The Conda runtime is another good option for HPC, I think, though not as isolated as the containerization provided by Singularity.

j23414:
> The Conda runtime is another good option for HPC, I think, though not as isolated as the containerization provided by Singularity.

Nod, or this [this workaround](https://github.com/nextstrain/cli/issues/2#issuecomment-599172854)

tsibley:
To test this, the easiest way is to download the standalone archive built by CI and available as a CI artifact.

Artifacts can only be downloaded by authenticated users per GitHub's policy, so you'll either need to download the archive in your browser:

https://github.com/nextstrain/cli/suites/10229277255/artifacts/501436118

or have credentials in your `~/.netrc` for an authenticated `curl` invocation:

    curl -fsSL --netrc https://api.github.com/repos/nextstrain/cli/actions/artifacts/501436118/zip > standalone-x86_64-unknown-linux-gnu.zip

Once you have `standalone-x86_64-unknown-linux-gnu.zip`, then unpack it and setup the runtime:

```
mkdir -p cli-test
funzip < standalone-x86_64-unknown-linux-gnu.zip | tar xzvp -C cli-test
./cli-test/nextstrain setup singularity
```

tsibley:
> Nod, or this [this workaround](https://github.com/nextstrain/cli/issues/2#issuecomment-599172854)

Sure, though this runtime would obsolete that workaround…

tsibley:
New, simpler way to install the standalone build for this PR:

```
curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux \
    | DESTINATION=/tmp/cli-test bash -s ci-build/3859193828
```

(Download will be much slower than a normal release, but that's ok for dev.)

No need to run the two final commands the installer suggests at the end.

Then get started testing it with:

```
/tmp/cli-test/nextstrain setup singularity
```

tsibley:
Rebased onto latest `master` branch to account for #250.

tsibley:
[Per Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1674066213797409), I'm going to merge this today unless there are objections. It's worked in my testing, incl. on the Hutch's cluster. And I think @j23414 reported it worked for her too. I'm sure there will be things to fix that come up as we use it, so good to get it out there and available. I don't intend to add it yet to our main installation docs, but it will be noted in the Nextstrain CLI docs, and we can offer it in situations it might be useful.

PR #250, opened 2023-01-13
Doc updates

tsibley:
A bit of a grab bag. See commit messages.

### Testing

- [x] Doc preview looks good
- [x] Checks pass


doc/installation.rst
```diff
@@ -0,0 +1,211 @@
+============
```
victorlin, line 1: _(non-blocking)_

Nice pandoc command in 1272b0ea00e5d90e87c7f14b6a6d3c6291685d8b. I like it better than what I had done in https://github.com/nextstrain/docs.nextstrain.org/commit/f97ae0c30d6ecb4377b321bbdc4d917563d424e7 and other previous conversions.

doc/glossary.rst
```diff
@@ -0,0 +1,52 @@
+========
+Glossary
+========
+
+Terms used by Nextstrain CLI.  Primarily intended for developers new to the
```
victorlin, line 5: _(non-blocking)_

"runner" / "runtime" / "computing platform" is a nice way to describe the different pieces 👌

Defining and standardizing these terms enables us to better communicate how things work both for users and ourselves.

victorlin (approved):

doc/glossary.rst
tsibley: :raised_hands:

PR #252, opened 2023-02-06
standalone-installer-unix: Check for Rosetta 2 in a way that works on macOS 11 too

tsibley:
The previous check worked only on macOS 12 and 13, because the executable path on macOS 11 is:

    /Library/Apple/usr/libexec/oah/oahd

On all macOS versions, however, the process runs as the _oahd user.  In case the _oahd user doesn't exist, direct stderr to the bit bucket to avoid a noisy error being confusing to the person running the installer.

Troubleshooting happened on Slack¹ with the help of @cassiawag and @corneliusroemer.

¹ <https://bedfordlab.slack.com/archives/C0K3GS3J8/p1675714639052449>

### Testing

- [ ] Checks pass

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -115,7 +115,7 @@ target-triple() {
 
         Darwin)
             [[ "$machine" == x86_64 || "$machine" == arm64 ]] || die "unsupported architecture: $machine"
-            [[ "$machine" != arm64 ]] || pgrep -qf /usr/libexec/rosetta/oahd || die "Rosetta 2 not enabled.  Please run:"$'\n\n'"    softwareupdate --install-rosetta"$'\n\n'"and then retry this installation of Nextstrain CLI."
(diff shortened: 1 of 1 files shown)


tsibley:
Merging because all the standalone installer tests in CI passed, and it works in my own testing on macOS 12. Worst case, I think, is that it _doesn't_ work on macOS 11 arm64 (but _does_, in CI, on macOS 11 x86_64), and I have to fix it again. But that case doesn't change the situation of it not working on macOS 11 arm64 now anyway, so low risk.

tsibley:
@cassiawag [reports](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1675722966615549?thread_ts=1675714639.052449&cid=C0K3GS3J8) the fix does work on 11.7.3.

corneliusroemer:
Works on macOS 13.2 as well

PR #253, opened 2023-02-17
[batch] Improvements when attaching to completed jobs

tsibley:

### Testing

- [x] Manual testing of behaviour locally
- [x] Checks pass

### DIFF
--- a/nextstrain/cli/command/build.py
+++ b/nextstrain/cli/command/build.py
@@ -98,6 +98,22 @@ def register_parser(subparser):
         dest   = "download",
         action = "store_false")
 
+    # A --logs option doesn't make much sense right now for most of our
+    # runtimes, but I can see how it might in the future.  So we're ready if
+    # that future comes to pass, set up --no-logs as if there's a --logs option
+    # enabled by default.  This also avoids a double negative in conditions,
+    # e.g. avoids writing "if not opts.no_logs".
+    #   -trs, 9 Feb 2023
+    parser.set_defaults(logs = True)
+    parser.add_argument(
+        "--no-logs",
+        help   = "Do not show the log messages of the remote build. "
+                 "Currently only supported when also using --aws-batch. "
+                 "Default is to show all log messages, even when attaching to a completed build."
+                  f"{SKIP_AUTO_DEFAULT_IN_HELP}",
+        dest   = "logs",
+        action = "store_false")
+
     # Positional parameters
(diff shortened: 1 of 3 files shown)


nextstrain/cli/command/build.py
```diff
@@ -98,6 +98,22 @@ def register_parser(subparser):
...
+    #   -trs, 9 Feb 2023
+    parser.set_defaults(logs = True)
+    parser.add_argument(
+        "--no-logs",
+        help   = "Do not show the log messages of the remote build. "
+                 "Currently only supported when also using --aws-batch. "
+                 "Default is to show all log messages, even when attaching to a completed build."
+                  f"{SKIP_AUTO_DEFAULT_IN_HELP}",
+        dest   = "logs",
+        action = "store_false")
```
joverlee521, line 115: I'm confused by the use of `set_defaults` here. I thought using `store_false` will already set the default value to `True`?

joverlee521 (approved):
Changes LGTM. 

The tests for Python 3.6 are failing...do we still need to support 3.6?

nextstrain/cli/command/build.py
corneliusroemer: It's btw not the PR that's causing tests to fail - CI fails also on master: https://github.com/nextstrain/cli/actions/runs/4266385449

I've made a PR to bump minimum supported Python version to 3.7: #256

corneliusroemer:
>The tests for Python 3.6 are failing...do we still need to support 3.6?

It's btw not the PR that's causing tests to fail - CI fails also on master: https://github.com/nextstrain/cli/actions/runs/4266385449

I've made a PR to bump minimum supported Python version to 3.7: https://github.com/nextstrain/cli/pull/256

nextstrain/cli/command/build.py
tsibley: Ah, it will, but I found it clearer to be explicit given that the destination for `--no-logs` is `logs`. If you think it's much more confusing, I can remove the `set_defaults` call.

tsibley:
Rebased onto latest master to pick up fixes for CI (and fix changelog conflicts).

nextstrain/cli/command/build.py
joverlee521: Ah, no need to remove it. Just wanted to make sure I wasn't missing something.

PR #261, opened 2023-03-01
botocore 1.28.0 compat

tsibley:

### Testing

- [x] Confirmed the issue is resolved by this change
- [x] Checks pass

### DIFF
--- a/nextstrain/cli/runner/aws_batch/jobs.py
+++ b/nextstrain/cli/runner/aws_batch/jobs.py
@@ -288,9 +288,11 @@ def override_definition(base_definition_name: str, image: str) -> str:
         derived_definition["jobDefinitionName"] = derived_definition_name
         derived_definition["containerProperties"]["image"] = image
 
-        # These are AWS-assigned keys returned by describe_job_definitions() which
+        # Remove AWS-assigned keys returned by describe_job_definitions() which
         # aren't supported as keyword arguments by register_job_definition().
-        for key in {'jobDefinitionArn', 'revision', 'status', 'containerOrchestrationType'}:
+        register_inputs = batch.meta.service_model.operation_model("RegisterJobDefinition").input_shape.members
+
+        for key in set(derived_definition) - set(register_inputs):
             del derived_definition[key]
 
         batch.register_job_definition(**derived_definition)
--- a/nextstrain/cli/runner/aws_batch/jobs.py
+++ b/nextstrain/cli/runner/aws_batch/jobs.py
@@ -290,7 +290,7 @@ def override_definition(base_definition_name: str, image: str) -> str:
 
         # These are AWS-assigned keys returned by describe_job_definitions() which
         # aren't supported as keyword arguments by register_job_definition().
-        for key in {'jobDefinitionArn', 'revision', 'status'}:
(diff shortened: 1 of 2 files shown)


nextstrain/cli/runner/aws_batch/jobs.py
```diff
@@ -288,10 +288,13 @@ def override_definition(base_definition_name: str, image: str) -> str:
         derived_definition["jobDefinitionName"] = derived_definition_name
         derived_definition["containerProperties"]["image"] = image
 
-        # These are AWS-assigned keys returned by describe_job_definitions() which
+        # Remove AWS-assigned keys returned by describe_job_definitions() which
         # aren't supported as keyword arguments by register_job_definition().
-        for key in {'jobDefinitionArn', 'revision', 'status'}:
-            del derived_definition[key]
+        register_inputs = set(batch.meta.service_model.operation_model("RegisterJobDefinition").input_shape.members)
```
huddlej, line 293: This looks like an awesome direction! Are all of the attributes/methods access through botocore's here part of their public API? I don't know anything about botocore and how it wraps AWS APIs, but I didn't immediately find `operation_model` in [the Batch API docs](https://boto3.amazonaws.com/v1/documentation/api/1.12.21/reference/services/batch.html), for example.

joverlee521: ☝️ I have the same question. How did you find such a deeply nested attribute?

tsibley: The `batch` client object here is part of botocore, not boto3, even though we construct it via boto3. The [botocore docs](https://botocore.amazonaws.com/v1/documentation/api/latest/index.html#), which are almost entirely generated, don't cover the service meta and model stuff, so I'm not sure what the public vs. private nature is. The docs don't explicitly address public vs. private API surfaces. However, the attributes accessed here seem at least semi-public to me, as there are many private-by-convention attributes starting with a leading underscores.

I found the attribute chain thru inspection, knowing that they'd likely be there as botocore had data-defined models of all the services and API endpoints and used those to generate and drive its client code. (If the attributes weren't there, I was prepared to load and traverse the JSON files [defining the models](https://github.com/boto/botocore/blob/567cb9da384b4be872389480d92aac8550b9ec94/botocore/data/batch/2016-08-10/service-2.json#L239-L252) directly, but that would be more brittle, I think.)

joverlee521: Ohh, I see. botocore uses [these attributes to generate their own docs](https://github.com/boto/botocore/blob/1fb4d3a69a46941aee66314b0983d316659cbc60/botocore/docs/method.py#L67).

joverlee521 (approved):

nextstrain/cli/runner/aws_batch/jobs.py
tsibley: Yeah. My guess is that AWS would disclaim support for these attributes and consider them only public within botocore's own internals, but I think they're likely to be stable enough to be ok for us to use. (That is, at least as stable as hardcoding a list of fields.)

tsibley:
Repushed with a small but I think useful code simplification. Testing locally it still works as expected.

PR #262, opened 2023-03-07
standalone-installer-unix: Use `eval` instead of `source` to avoid issues on old versions of Bash

tsibley:
In Bash prior to 4.0 (released in 2009), the `source` builtin does not properly read input from a pipe, named or otherwise.  When given a pipe, such as via process substitution (i.e. `<(…)`), it silently does nothing and exits 0.  This can cause a SIGPIPE in the process on other end of the pipe and, in Python, a BrokenPipeError.

Unfortunately, macOS ships with Bash 3.2.57 (released in 2007) and, for legal reasons, is unlikely to ever be upgraded.  Switching to `eval` with command substitution (i.e. `$(…)`) avoids the issue for Bash users on macOS and results in no functional change for everyone else.

We _will_ have to ensure that the output of `nextstrain init-shell` remains below command line length limits since the output is now interpolated into the command line, but I don't think this will ever be an issue since the initialization sequence should remain pretty short.

Initial problem report and troubleshooting with @alliblk and @victorlin happened in Slack.¹

¹ <https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1677871789858519>

### Testing

- [x] Tested locally in my bashrc and interactively
- [x] Checks pass

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -72,7 +72,7 @@ main() {
     log "Checking installation"
     local installed_version
     if ! installed_version="$("$DESTINATION"/nextstrain --version)"; then
-        die "installation check failed: unable to run \`"$DESTINATION"/nextstrain --version\`; look for error message above."
+        die "installation check failed: unable to run \`\"$DESTINATION\"/nextstrain --version\`; look for error message above."
     fi
 
     cat <<~~
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -91,8 +91,8 @@ Nextstrain CLI ($installed_version) installed to $DESTINATION.
 To make the "nextstrain" command available in your default shell ($shell)
 without using the full path, please run these two commands now:
 
-    echo 'source <("$DESTINATION/nextstrain" init-shell $shell)' >> $rc
-    source <("$DESTINATION/nextstrain" init-shell $shell)
+    echo 'eval "$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
+    eval "$("$DESTINATION/nextstrain" init-shell $shell)"
 
 The first adds a line to your shell initialization file ($rc) for future
 sessions.  The second sets up your current shell session.  You only need to run
--- a/nextstrain/cli/command/init_shell.py
+++ b/nextstrain/cli/command/init_shell.py
@@ -6,7 +6,7 @@ If PATH does not contain the expected installation path, emits an appropriate
 
 Use this command in your shell config with a line like the following::
 
-    source <({INSTALLATION_PATH}/nextstrain init-shell)
+    eval "$({INSTALLATION_PATH}/nextstrain init-shell)"
 
 Exits with error if run in an non-standalone installation.
 """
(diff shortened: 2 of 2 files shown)


victorlin:
> We _will_ have to ensure that the output of `nextstrain init-shell` remains below command line length limits since the output is now interpolated into the command line, but I don't think this will ever be an issue since the initialization sequence should remain pretty short.

Is "command line length limits" the output of `getconf ARG_MAX`, which varies by OS?

I trust your judgement that this won't ever be an issue, but in the small chance that it does, this would likely only be caught by user complaint. Could a check be added in CI, or at least a note in the source code of `nextstrain init-shell` to be wary of how the output is used?

tsibley:
> Is "command line length limits" the output of `getconf ARG_MAX`, which varies by OS?

Not just `getconf ARG_MAX`, but yes, varies by OS and by system.

> I trust your judgement that this won't ever be an issue, but in the small chance that it does, this would likely only be caught by user complaint. Could a check be added in CI, or at least a note in the source code of `nextstrain init-shell` to be wary of how the output is used?

Yeah, good idea. I've added a comment and repushed.

victorlin (approved):

victorlin:
So I approved this without actually testing it, and I just happened to run the installer for something else... a proper test now shows something wrong here.

This is what my console shows:

```
% curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac | bash
--> Temporary working directory: /var/folders/rk/s2p6nv1d13lcf0dynjkvxyl80000gp/T/tmp.Ms3Wnwnj
--> Downloading https://nextstrain.org/cli/download/latest/standalone-x86_64-apple-darwin.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   138  100   138    0     0    335      0 --:--:-- --:--:-- --:--:--   338
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 34.9M  100 34.9M    0     0  1779k      0  0:00:20  0:00:20 --:--:-- 1706k
--> Extracting standalone-x86_64-apple-darwin.tar.gz
--> Removing existing /Users/vlin/.nextstrain/cli-standalone
--> Installing to /Users/vlin/.nextstrain/cli-standalone
--> Cleaning up
--> Checking installation
______________________________________________________________________________

Nextstrain CLI (nextstrain.cli 6.2.0) installed to /Users/vlin/.nextstrain/cli-standalone.

To make the "nextstrain" command available in your default shell (zsh)
without using the full path, please run these two commands now:

    echo 'eval "# This will mask /opt/homebrew/Caskroom/miniconda/base/envs/nextstrain-docker/bin/nextstrain

# Add /Users/vlin/.nextstrain/cli-standalone to front of PATH
export PATH=/Users/vlin/.nextstrain/cli-standalone"${PATH:+:}$PATH""' >> ~/.zshrc
    eval "# This will mask /opt/homebrew/Caskroom/miniconda/base/envs/nextstrain-docker/bin/nextstrain

# Add /Users/vlin/.nextstrain/cli-standalone to front of PATH
export PATH=/Users/vlin/.nextstrain/cli-standalone"${PATH:+:}$PATH""

The first adds a line to your shell initialization file (~/.zshrc) for future
sessions.  The second sets up your current shell session.  You only need to run
these once.
```

It looks like `$("$DESTINATION/nextstrain" init-shell $shell)` gets evaluated for the output. Some characters might need escaping?

tsibley:
Yeah, I just noticed this myself locally and pushed bb7c149efacee28e4dbd843e54f5f7e17505ee0c and came here to comment about it. orz orz orz

PR #275, opened 2023-04-28
Add newline before eval command

victorlin:

Previously, the command could result in unexpected errors if the shell init file did not end with a newline (for reasons beyond our control).

If the file already ends with a newline, this comes with the added benefit of using whitespace to distinguish the "scope" of this command as separate from whatever was there before.

### Related issue(s)

I noticed this because my own `.zshrc` file did not end with a newline and I got this error upon opening new sessions after running the command:

```
/Users/vlin/.zshrc:63: parse error near `\n'
```

which made sense after inspecting the file contents:

```
function some-function()
{
    …
}eval "$("/Users/vlin/.nextstrain/cli-standalone/nextstrain" init-shell zsh)"
```

### Testing

- [x] Ran `bin/standalone-installer-unix` locally to verify output can be copy-pasted as-is
- [x] Standalone installers checks pass, outputs look good there too
- [x] ~Other checks pass~ failing because of unrelated reasons, but this change should not affect the outcome of those checks

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -92,7 +92,7 @@ Nextstrain CLI ($installed_version) installed to $DESTINATION.
 To make the "nextstrain" command available in your default shell ($shell)
 without using the full path, please run these two commands now:
 
-    echo 'eval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
+    printf '\n%s\n' 'eval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
     eval "\$("$DESTINATION/nextstrain" init-shell $shell)"
 
 The first adds a line to your shell initialization file ($rc) for future


bin/standalone-installer-unix
```diff
@@ -92,7 +92,7 @@ Nextstrain CLI ($installed_version) installed to $DESTINATION.
 To make the "nextstrain" command available in your default shell ($shell)
 without using the full path, please run these two commands now:
 
-    echo 'eval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
+    echo '\neval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
```
tsibley, line 95: bash (and many other shells) won't expand `\n` within this string (either at expansion time or within `echo`).

I do see that zsh will (at least by default).

We can use `echo -e` instead as long as we ensure we don't use any other recognized escapes (we currently don't).

Or we can switch to something that doesn't mix an interpreted string with uninterpreted data, like

```
printf '\n%s\n' 'eval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> $rc
```

Another alternative is using the syntax for an interpreted string literal and string concat:

    echo $'\n''eval "\$("$DESTINATION/nextstrain" init-shell $shell)"'

And if we do that, for consistency across shells we can also throw in `-E` to avoid `echo` interpreting the string.

tsibley (changes requested):
Unix has a very strong convention that files end in newlines, but still good to avoid this issue I guess. We do need to tweak the proposed change, however, so it works on both bash and zsh (at the least).

bin/standalone-installer-unix
tsibley: > Use printf to avoid expanding '\n' in bash, zsh, and hopefully all other
shells.

Ah, not to avoid expanding `\n` but to consistently expand it.

tsibley (approved):

bin/standalone-installer-unix
victorlin: Oops, I got mixed up. Fixed in ac5d20e21c3ed2ca3cd23cc207f164a10ff73efb

Issue #274, opened 2023-04-25
Read-only filesystem errors from Snakemake in Singularity runtime


tsibley:
The Snakemake upgrade in https://github.com/nextstrain/docker-base/pull/136 caused failures for builds in the Singularity runtime due to new filesystem writes that Snakemake now tries to make under `$HOME`. Currently with our Singularity runner, `$HOME` and other non-working dirs are read-only.

The failure showed up as scheduled daily CI in this repo starting to fail between `nextstrain/base:build-20230407T002437Z` ([last good](https://github.com/nextstrain/cli/actions/runs/4659776212/jobs/8247053194)) and `build-20230411T103027Z` ([first bad](https://github.com/nextstrain/cli/actions/runs/4670446582/jobs/8270193167)).

I think we can address this by using the `--writable-tmpfs` flag in the Singularity runner to allow a writable (but temporary and ultimately discarded) filesystem outside of the working dir. This is similar to Docker's default behaviour. There may be other considerations with the change though.

tsibley:
Maybe (or maybe not) related to this is that the `nextstrain shell` prompt also no longer shows up for the Singularity runtime. Instead, we get the default `Singularity>` shell prompt. But this might be Singularity version dependent also.

tsibley:
I keep getting sidetracked on this. I'll pick it back up in a week or so, unless someone else gets to it. My extremely "for my eyes" notes on this:

```diff
diff --git a/nextstrain/cli/resources/bashrc b/nextstrain/cli/resources/bashrc
index 3456696..83a83a4 100644
--- a/nextstrain/cli/resources/bashrc
+++ b/nextstrain/cli/resources/bashrc
@@ -6,8 +6,13 @@ fi
 
 __NEXTSTRAIN_BASHRC=1
 
+OLDHOME="${HOME:-}"
+HOME=/nextstrain
+
 # Override PS1, this file's reason for being.
 OLDPS1="${PS1:-}"
+OLDPROMPT_COMMAND="${PROMPT_COMMAND:-}"
+unset PROMPT_COMMAND
 
 if [[ -n "${NEXTSTRAIN_PS1:-}" ]]; then
     PS1="$NEXTSTRAIN_PS1"
diff --git a/nextstrain/cli/runner/singularity.py b/nextstrain/cli/runner/singularity.py
index 9e7a671..3ba48cb 100644
--- a/nextstrain/cli/runner/singularity.py
+++ b/nextstrain/cli/runner/singularity.py
[...]

PR #277, opened 2023-04-28
standalone-installer-windows: Use call operator (&) instead of Invoke-Expression

### LINKED ISSUE #276: Installation fails when there is a space in the path
Originally from a [discussion post](https://discussion.nextstrain.org/t/trouble-installing-nextstrain-cli/1369).

### Current Behavior

The standalone installation command fails when there is a space in the installation path, for example in the user's home directory:

asdf

### Expected behavior

Installation happens without failure.

### Possible solution

[Use `&` instead of `Invoke-Expression`](https://devblogs.microsoft.com/powershell/invoke-expression-considered-harmful/).

victorlin:

It is not uncommon for user home directories to have spaces, and Invoke-Expression [does not work well](https://devblogs.microsoft.com/powershell/invoke-expression-considered-harmful/) with paths that have spaces. The call operator is more suitable for this use case.

Quoting updated to reflect [call operator](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.3#call-operator-) [syntax](https://ss64.com/ps/call.html).

### Related issue(s)

Fixes #276 

### Testing

- [x] Manually tested on a Windows VM installing to `C:\Users\test user\`.
- [x] ~Checks pass~ some are failing because of unrelated reasons, but this change should not affect the outcome of those checks

### DIFF
--- a/bin/standalone-installer-windows
+++ b/bin/standalone-installer-windows
@@ -108,7 +108,7 @@ function main([string]$version) {
         Remove-Item -Recurse -Force $tmp
     }
 
-    $version = Invoke-Expression "$destination\nextstrain --version"
+    $version = & "$destination\nextstrain" --version
 
     echo @"
 ______________________________________________________________________________


tsibley (approved):
Ah, yes, this is much better. I vaguely knew about PowerShell's call operator, but didn't really grok the subtlety until now.

Totally minor, but in the commit message, I'd probably link to the official docs—<https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.3#call-operator->—instead of a third-party site.

CI is failing due to #274.

victorlin:
> in the commit message, I'd probably link to the official docs … instead of a third-party site.

Fair point. Updated with both links, since the official docs don't provide the syntax reference.

PR #278, opened 2023-05-03
Standalone installer parity

victorlin:

Currently, the UNIX and Windows installers are slightly different. These changes make them less different.

### Testing

- [x] Ran `bin/standalone-installer-unix` locally on macOS with `zsh`
- [ ] Standalone installers checks pass, outputs look good there too
- [x] ~Other checks pass~ failing because of unrelated reasons, but this change should not affect the outcome of those checks

### DIFF
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -71,11 +71,11 @@ main() {
         rc="$(shell-rc "$shell")"
 
         # Initialize for this session.
-        eval "$("$DESTINATION/nextstrain" init-shell $shell)"
+        eval "$("$DESTINATION/nextstrain" init-shell "$shell")"
 
         # Automatically initialize for new sessions.
         # shellcheck disable=SC2016
-        printf '\n%s\n' 'eval "\$("$DESTINATION/nextstrain" init-shell $shell)"' >> "$rc"
+        printf '\n%s\n' 'eval "\$("$DESTINATION/nextstrain" init-shell "$shell")"' >> "$rc"
     fi
 
     popd >/dev/null
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -87,8 +87,8 @@ main() {
 
     log "Checking installation"
     local installed_version
-    if ! installed_version="$("$DESTINATION"/nextstrain --version)"; then
-        die "installation check failed: unable to run \`\"$DESTINATION\"/nextstrain --version\`; look for error message above."
+    if ! installed_version="$(nextstrain --version)"; then
+        die "installation check failed: unable to run \`nextstrain --version\`; look for error message above."
     fi
 
     cat <<~~
--- a/bin/standalone-installer-unix
+++ b/bin/standalone-installer-unix
@@ -63,6 +63,21 @@ main() {
     mkdir -p "$(dirname "$DESTINATION")"
     mv standalone "$DESTINATION"
 
+    if [[ "$(type -p nextstrain 2>/dev/null)" != "$DESTINATION/nextstrain" ]]; then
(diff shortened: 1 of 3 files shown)


bin/standalone-installer-windows
victorlin, line 1: 08ca903e8d972c9afd857ecf3e205ab2fc33da26 has a merge conflict with 6f89bf438b87014082afe6aa7b939b0de5adf858. This PR is currently based on #275 so I'll wait to rebase and fix the merge conflict until after that PR is merged.

bin/standalone-installer-unix
victorlin, line 1: @tsibley heads up that this is still draft - mainly I haven't finalized f3ecf314201b090c017c5036504618147b912a39, which is something that was brought up with https://github.com/nextstrain/cli/pull/217#issuecomment-1230851374 and ultimately not implemented, but I want to give it a shot here.

tsibley: Yep! I know. :-) I noticed the activity and so quickly skimmed thru, and in doing so saw a few things that I'm :thinking: about. So I flagged myself for review to make sure to put :eyes: back on this once it's ready. (Didn't want to jump in and make a bunch of comments on something you hadn't flagged as "ready"!)

victorlin: Got it! It's really only f3ecf314201b090c017c5036504618147b912a39 that I need to fix up, so if you have comments on other commits feel free to send those in.

tsibley: I think my biggest objection is that I don't think we should be modifying people's shell config files automatically for them...

bin/standalone-installer-unix
```diff
@@ -63,6 +63,21 @@ main() {
...
 
+    if [[ "$(type -p nextstrain 2>/dev/null)" != "$DESTINATION/nextstrain" ]]; then
+        log "Making Nextstrain CLI available for the current user"
+
+        local shell rc
+        shell="$(default-shell)"
+        rc="$(shell-rc "$shell")"
+
+        # Initialize for this session.
+        eval "$("$DESTINATION/nextstrain" init-shell $shell)"
```
tsibley: You may know this, but this won't affect the calling shell session, just the session this program is running within.

bin/standalone-installer-unix
```diff
@@ -87,8 +87,8 @@ main() {
 
     log "Checking installation"
     local installed_version
-    if ! installed_version="$("$DESTINATION"/nextstrain --version)"; then
-        die "installation check failed: unable to run \`\"$DESTINATION\"/nextstrain --version\`; look for error message above."
+    if ! installed_version="$(nextstrain --version)"; then
+        die "installation check failed: unable to run \`nextstrain --version\`; look for error message above."
```
tsibley, line 91: The use of the full path was intentional, as it won't ever be misleading if the PATH lookup doesn't go as expected (e.g. because shell setup above didn't fail but didn't work). I don't think we should use a PATH lookup here.

tsibley:
af821c0b367b612b59b316c7ffcf984efe8ad164 seems like a useful edge case to address, but I'm skeptical of the direction the other commits on this branch take. I'm going to close this PR for now and we can revisit later if the need/desire arises.

@victorlin If you disagree, please feel free to re-open. (But I'd also hope if we re-open to then push it to a conclusion.)

tsibley:
Ah! and we'd already extracted af821c0b367b612b59b316c7ffcf984efe8ad164 from this branch into ac5d20e21c3ed2ca3cd23cc207f164a10ff73efb.

victorlin:
Agreed, thanks for closing this out!

PR #281, opened 2023-05-13
Fix Pyright errors

joverlee521:

Fix Pyright errors that have been raised in our scheduled CI runs since the release of Pyright 1.1.306.¹ I have not dug into the Pyright changelog to determine the exact change that leads to this new error.

¹ https://github.com/nextstrain/cli/actions/runs/4874920417

### Testing

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/runner/aws_batch/s3.py
+++ b/nextstrain/cli/runner/aws_batch/s3.py
@@ -11,7 +11,7 @@ from calendar import timegm
 from os import utime
 from pathlib import Path
 from time import struct_time
-from typing import Callable, Generator, Iterable, List, Optional
+from typing import Callable, Generator, Iterable, List, Optional, Any
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


nextstrain/cli/runner/aws_batch/s3.py
```diff
@@ -65,6 +65,7 @@ def upload_workdir(workdir: Path, bucket: S3Bucket, run_id: str) -> S3Object:
     ])
 
     # Stream writes directly to the remote ZIP file
+    remote_file: Any
```
joverlee521, line 68: In the future, we may consider creating [type stubs](https://github.com/microsoft/pyright/blob/main/docs/type-stubs.md) for `fsspec` so that we can provide the correct type instead of `Any`.

joverlee521:
Merging since this is a small type fix that will only affect the Pyright test. The checks that are currently failing are the ones related to https://github.com/nextstrain/cli/issues/274

PR #287, opened 2023-06-01
Revert "Require docutils <0.16 in dev deps to avoid a version conflict"

tsibley:
This reverts commit d2463b72c153ea6720d920baff01003d51ec4c04.

Pip's dependency resolution around docutils no longer appears to be an issue.  We also no longer test on Python 3.6.1 in CI but instead use 3.6.15.

This pin was causing issues in CI tests itself since it was preventing use of newer docutils with fixes for deprecated pkg_resource APIs.  The older docutils versions caused warnings at test time.

Since docutils is also a non-dev dependency and unpinned there, it's also nice to use the same version in dev/test as in non-dev installs.

### Testing

- [x] Tests pass locally
- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,8 +62,7 @@ jobs:
       - name: Install Nextstrain CLI
         run: python3 -m pip install --upgrade '.[dev]'
 
-      - name: Run pytest
-        run: pytest -v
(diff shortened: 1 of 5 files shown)


joverlee521 (approved):
Still need to familiarize myself more with the testing in this repo, but the changes look to me.

PR #289, opened 2023-06-12
Environment variables

tsibley:
See commit messages.

Other context in my notes:

- [26 Aug 2022](https://github.com/tsibley/blab-standup/blob/master/2022-08-26.md#26-august-2022)
- [2 June 2023](https://github.com/tsibley/blab-standup/blob/master/2023-06-02.md#2-june-2023)

### Testing

- [x] Checks pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,57 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Improvements
+
+* Commands that use a runtime (`nextstrain build`, `nextstrain shell`, and
+  `nextstrain view`) now support two new options for setting or passing thru
+  environment variables into the runtime environment:
+
+      --env <name>[=<value>]
+      --envdir <path>
+
+  When either of these options are given, the default behaviour of
+  automatically passing thru several "well-known" environment variables is
+  disabled.  That is, the following "well-known" environment variables are only
+  automatically passed thru when the new `--env` and `--envdir` options are
+  _not_ used:
+
+    - `AUGUR_RECURSION_LIMIT`
+    - `AUGUR_MINIFY_JSON`
+    - `AWS_ACCESS_KEY_ID`
+    - `AWS_SECRET_ACCESS_KEY`
+    - `AWS_SESSION_TOKEN`
+    - `ID3C_URL`
+    - `ID3C_USERNAME`
+    - `ID3C_PASSWORD`
+    - `RETHINK_HOST`
+    - `RETHINK_AUTH_KEY`
+
+  Pass these variables explicitly via `--env` or `--envdir` if you need them in
+  combination with other `--env` or `--envdir` usage.  For more usage details,
+  use the `--help-all` flag of any of those commands, e.g. `nextstrain build
+  --help-all`.
+  ([#289](https://github.com/nextstrain/cli/pull/289))
+
+* Environment variables are now passed to the Docker and AWS Batch runtimes via
+  more secure means when the container image in use is new enough to support it
+  (`nextstrain/base:build-20230613T204512Z` and newer).  This ensures the env
+  values aren't visible in the container's config (e.g. via `docker inspect`,
+  `aws batch describe-jobs`, the AWS web console).  If you're using an older
+  image, you can update it with `nextstrain update docker`.
+
+  For Docker, environment variables are written to an internal and temporary
+  envdir directory visible only to the current user which is deleted
+  immediately after use at container start.
+
+  For AWS Batch, environment variables are written to a ZIP archive on S3,
+  alongside but separate from the ZIP archive of the build dir.  This env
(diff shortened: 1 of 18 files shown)


tsibley:
Looks like CI caught an issue on Python 3.8 and up in some new doctests that don't explicitly cleanup a temp dir:

    ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmpc8f2f210'>

This warning is turned into an error by our pytest config, and only happens because of some doctest code anyway, so it's a CI-specific failure. I'll address it nevertheless.

Hilariously, the warning-turned-error can't be raised, presumably because the warning is emitted during global destruction, and this causes _another_ warning, `pytest.PytestUnraisableExceptionWarning`.

tsibley:
A couple other CI issues cropping up. Not sure why they're happening now, but I'll dig. Seem unrelated to this PR.

- [macOS with Python 3.7 failed](https://github.com/nextstrain/cli/actions/runs/5294615024/jobs/9584149563) (and only that one!) because of missing bz2 support in Python, which is optional but assumed to be present by fsspec, which is in the import chain of most/all of our code here, so everything falls to pieces. I think the missing bz2 support is some transient/temporary issue with the Python we're getting or the runner or something and we should fix/workaround that, but fsspec should also not assume bz2 support, so I filed a  [fix for that](https://github.com/fsspec/filesystem_spec/pull/1295).
- All of the test-source jobs on Windows failed because Windows lets you read a directory like a file (it appears empty) and thus a doctest in the envdir reading code failed because it expects a `IsADirectoryError` instead. TIL!

tsibley:
CI now passes! I addressed the Windows issue by checking for directories explicitly so it's compat across platforms, and the macOS bz2 issue was apparently indeed transient.

tsibley:
One thing I want to consider (and think we should do) to help deprecate and remove the automatic hostenv forwarding is to disable it when these new `--env` or `--envdir` options are used. Since they're new, their use is a good time/indicator to push folks to spell out what they need without hostenv. Thoughts?

tsibley:
I've done that now in https://github.com/nextstrain/cli/pull/289/commits/fa2633dc3f832ac8ff1d957147efee9a7b31972e.

tsibley:
CI failures appear unrelated: one of the Markdown image inlining tests is [failing](https://github.com/nextstrain/cli/actions/runs/5314773504/jobs/9622365358#step:5:89) and started failing with yesterday's scheduled CI. I assume some new release of a dep broke something.

tsibley:
Rebased onto latest master after merge of #294.

LICENSE.id3c
```diff
@@ -0,0 +1,26 @@
+This license applies to the original copy of "prose_list" from the ID3C project
+into this project, incorporated as part of "nextstrain/cli/util.py".  Any
+subsequent modifications this project's copy of "prose_list" are licensed under
```
victorlin, line 3:
```suggestion
subsequent modifications to this project's copy of "prose_list" are licensed under
```

CHANGES.md
```diff
@@ -13,6 +13,57 @@ development source code and as such may not be routinely kept up to date.
...
+  environment variables into the runtime environment:
+
+      --env <name>[=<value>]
+      --envdir <path>
+
+  When either of these options are given, the default behaviour of
+  automatically passing thru several "well-known" environment variables is
+  disabled.  That is, the following "well-known" environment variables are only
+  automatically passed thru when the new `--env` and `--envdir` options are
+  _not_ used:
```
victorlin, line 29: > One thing I want to consider (and think we should do) to help deprecate and remove the automatic hostenv forwarding is to disable it when these new `--env` or `--envdir` options are used. Since they're new, their use is a good time/indicator to push folks to spell out what they need without hostenv. Thoughts?

Makes sense to me!

victorlin (approved):
I browsed through the commits and did a basic test using `nextstrain shell --env` and `nextstrain shell --envdir`. Good changes!

nextstrain/cli/runner/docker.py
```diff
@@ -29,6 +31,16 @@
 COMPONENTS = ["augur", "auspice", "fauna", "sacra"]
 
 
+class IMAGE_FEATURE(Enum):
+    # Auspice ≥1.35.2 necessitated tightly-coupled changes to the image (first
+    # present in this tag) and `nextstrain view` (see commit 2754026, first
+    # present in the 1.8.0 release).
+    compatible_auspice = "build-20190119T045444Z"
```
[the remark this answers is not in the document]
tsibley: I had to do some correlated history diving across repos to reconstruct the reason for this magic number. :upside_down_face:

LICENSE.id3c
tsibley: Ah, fixed up, thanks!

PR #290, opened 2023-06-12
ci: Fix silent failure to run pytest on Windows

tsibley:
The change from `pytest` to `./devel/pytest` in "dev: Set
NEXTSTRAIN_RST_STRICT=1 when running tests" (dc21d8d) caused pytest to
stop running entirely on our Windows jobs but not exit with an error
either.  The default shell there is pwsh (PowerShell) and it silently
(?!) fails to execute the Bash program, maybe because pwsh on Windows
doesn't support shebangs?

Explicitly use bash as the default shell for test-source jobs, similar
to the default we set for other jobs in this workflow.  This covers all
steps in the job.  Even though the "Install Nextstrain CLI" step works
on pwsh now, this guards against future modifications breaking it like I
inadvertently did with the pytest step.

Related-to: <https://github.com/nextstrain/cli/pull/287>

### Testing

- [x] Windows tests all run
- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,9 @@ jobs:
           - os: ubuntu-20.04
             python: '3.6'
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3


tsibley:
New failures… orz. In "test-dist" jobs, but only on Linux.

```
$ nextstrain version --verbose
  File "/home/runner/.local/bin/nextstrain", line 5, in <module>
    from nextstrain.cli.__main__ import main
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/__init__.py", line 20, in <module>
    from .command     import build, view, deploy, remote, shell, update, setup, check_setup, login, logout, whoami, version, init_shell, authorization, debugger
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/command/deploy.py", line 8, in <module>
    from .remote.upload import register_arguments, run, __doc__ # noqa: F401 (these are for re-export)
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/command/remote/__init__.py", line 34, in <module>
    from . import upload, download, ls, delete
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/command/remote/upload.py", line 24, in <module>
    from ...remote import parse_remote_path
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/remote/__init__.py", line 9, in <module>
    from . import (
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/remote/nextstrain_dot_org.py", line 77, in <module>
    from .. import markdown
  File "/home/runner/.local/lib/python3.10/site-packages/nextstrain/cli/markdown.py", line 35, in <module>
    from pyparsing import (
ImportError: cannot import name 'nested_expr' from 'pyparsing' (/usr/lib/python3/dist-packages/pyparsing.py)
```

tsibley:
Oh. Well, that output is from the "[test-dist (python=3.9 os=ubuntu-latest)](https://github.com/nextstrain/cli/actions/runs/5247569838/jobs/9477918010#logs)" job, but the traceback shows `python3.10`, so something's amiss.

tsibley:
Right, `shell: bash` meant that the Python provisioned with Conda by our custom "setup-integration-tests" action isn't found/used because that needs `bash -l …` so shell init files are sourced. When it was defaulting to `pwsh` on Windows this wasn't an issue since that always initializes Conda properly. orz

tsibley:
Oh. I missed this default for that job, so I didn't need to add `shell: bash` at all. Great.

https://github.com/nextstrain/cli/blob/a2559d1d6c7d5c7e92e66258dc6ce137de91a3d1/.github/workflows/ci.yaml#L232-L236

tsibley:
Alright, looks all better now.

joverlee521 (approved):
Nice job catching this! So weird that didn't show up as an error in the GitHub UI...I guess there's no error code if nothing ran? 

_non-blocking_

Would it make sense to just set the top level `defaults` to `bash`?

tsibley:
> So weird that didn't show up as an error in the GitHub UI...I guess there's no error code if nothing ran?

I think GitHub's not to blame here, but PowerShell. I think pwsh is failing to run the command (`./devel/pytest`) but not exiting non-zero, so GitHub's runner agent can't notice there was an error.

> Would it make sense to just set the top level `defaults` to `bash`?

I thought about it… but was vaguely worried it would cause more confusion rather than less. But maybe that's wrong! And maybe if it was our by-convention-default for workflows it'd be less confusing in the long run, e.g. compared to using the per-platform defaults.

One thing that's different between the default bash and setting `shell: bash`, for instance, is that with the former `-eo pipefail` is used but not with the latter.

joverlee521:
> maybe if it was our by-convention-default for workflows it'd be less confusing in the long run, e.g. compared to using the per-platform defaults.

Yeah, I think it'd be less confusing if we set a single default shell for our workflows. We can even use a custom set of shell options for our workflows instead of the built-in shell keywords.

PR #291, opened 2023-06-15
view: Wait for Auspice to start responding before opening the browser

tsibley:
A big improvement over simply pausing for 2s before opening it and hoping for the best.

### Testing

- [x] Manually, timing is right, and the timeouts work when I do `nextstrain view --exec cat /var/empty`
- [ ] Checks pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,13 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Bug fixes
+
+* `nextstrain view` now waits (up to 10s) for Auspice to start responding
+  before automatically opening it in the browser.  This should eliminate the
+  previous behaviour of sometimes opening the browser too soon.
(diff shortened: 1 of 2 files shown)


joverlee521 (approved):
The 2s timeout worked 50% of the time for me, so this is much better 😄

tsibley:
CI failure of a single job looks transient, but I'm re-running it now to see.

tsibley:
The non-changelog commit was successful after the re-attempt, so going to merge.

tsibley:
[Unrelated discussion of `nextstrain view` in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1694718932379359) led to me realizing that there's a failure mode here in merely waiting for the expected Auspice URL to start listening: if the port is already in use by another HTTP server (Auspice or otherwise), then the code will assume its the Auspice it launched from the current process and open it in the browser, even though in reality the launched Auspice errors out.

PR #292, opened 2023-06-16
runner.docker: Use more precise example commands in check-setup failure messages

tsibley:
The previous forms relied on defaults and outdated behaviour.  They predate the change to single-runtime application of `nextstrain update` from its previous application to "all runtimes, but in practice only Docker".  The more precise forms will be correct, for example, even if the Docker runtime isn't the default.

### Testing

- [x] `NEXTSTRAIN_DOCKER_IMAGE=nextstrain/base:build-20180119T045444Z nextstrain check-setup docker` shows new failure message and looks fine
- [x] Checks pass


tsibley:
Merging pre-review because this is a tiny change and CI passed.

PR #312, opened 2023-09-15
runner.conda: Explicitly install the latest package version during setup

tsibley:
This our intent and expectation, and it's good to be explicit about it. It may surface more installation issues, such as the one we observed in monkeypox CI¹ with the latest version not being installable, but obscuring those or surfacing them later on is _probably_ worse than addressing them head on earlier.  This change will mean that any macOS 10.14 users, if any, would have to use

    NEXTSTRAIN_CONDA_BASE_PACKAGE="nextstrain-base ==20230615T171309Z"

since newer versions aren't installable for them.²

This behaviour also parallels update's behaviour since "runner.conda: Explicitly specify a nextstrain-base version when updating" (d6e4f2b). It was an oversight (on my part) to not use the same behaviour during setup, but at the time I was focused on fixing an update bug.

Related-to: <https://github.com/nextstrain/conda-base/issues/41>
Related-to: <https://github.com/nextstrain/conda-base/pull/42>

¹ <https://github.com/nextstrain/monkeypox/issues/177>
² <https://github.com/nextstrain/conda-base/issues/38>

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/runner/conda.py
+++ b/nextstrain/cli/runner/conda.py
@@ -247,23 +247,30 @@ def setup_prefix(dry_run: bool = False, force: bool = False) -> bool:
         if not dry_run:
             shutil.rmtree(str(PREFIX))
 
-    # Conda packages to install.
+    # We accept a package match spec, which one to three space-separated parts.¹
+    # If we got a spec, then we use it as-is.
     #
-    # HEY YOU: If you add/remove packages here, make sure to account for how
-    # update() should make the same changes to existing envs.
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Confirmed that this no longer silently downgrades to `nextstrain-base ==20230615T171309Z` when I run `nextstrain setup conda --force` on macOS 10.14

tsibley:
Woo!

tsibley:
Ugh, forgot the changelog. Fixed in 1e8b12f.

PR #313, opened 2023-09-18
Avoid fsspec/s3fs 2023.9.1 due to errors caused by `auto_mkdir`

joverlee521:

Changes to the `auto_mkdir` parameter in https://github.com/fsspec/filesystem_spec/pull/1358 leads to the error described in https://github.com/fsspec/s3fs/issues/790.

Upstream fixes to come soon via https://github.com/fsspec/filesystem_spec/pull/1365 but good to avoid versions that we know are broken.

## Checklist

- [x] Checks pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@ development source code and as such may not be routinely kept up to date.
   view`.  They provide defaults when `--host` and/or `--port` aren't provided.
   ([#310](https://github.com/nextstrain/cli/pull/310))
 
+* Updated fsspec and s3fs dependencies to avoid version 2023.9.1, since they
+  caused `nextstrain build --aws-batch` invocations to fail.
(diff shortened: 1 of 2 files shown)


joverlee521:
Release with this change will fix our currently blocked automated AWS Batch builds that use the `setup-nextstrain-cli` action, which [still installs the Nextstrain CLI via pip](https://github.com/nextstrain/.github/blob/4d6afc489732204931ff527c27a17c3c505dc828/actions/setup-nextstrain-cli/action.yaml#L55).

tsibley (approved):
Looks good by inspection. Both the built wheel and standalone archive from the CI run get the expected fsspec/s3fs version for me locally and successfully submit AWS Batch jobs instead of being broken.

tsibley:
Ah, a brief changelog entry noting avoidance of a compatibility issue would be good.

Issue #319, opened 2023-09-26
Invalid `~/.shrc` file named in standalone installer instructions


tsibley:
When the default shell is `sh`, the standalone installer generates `~/.shrc` as the shell rc file to modify, which appears to be incorrect.

From @aperofsky [in Slack](https://bedfordlab.slack.com/archives/C036L7V615F/p1695705199036659):

> I have a Macbook Pro M1 Max running Ventura 13.6.
>
> At the end of the Nextstrain installation, there are the instructions: 
>
>     To make the "nextstrain" command available in your default shell (sh)
>     without using the full path, please run these two commands now:
>    
>       printf '\n%s\n' 'eval "$("/Users/.../.nextstrain/cli-standalone/nextstrain" init-shell sh)"' >> ~/.shrc
>       eval "$("/Users/.../.nextstrain/cli-standalone/nextstrain" init-shell sh)"
>    
>     The first adds a line to your shell initialization file (~/.shrc) for future
>     sessions. The second sets up your current shell session. You only need to run
>     these once.
>
> I needed to replace `~/.shrc` with `~/.profile` to get my default shell (sh) to recognize the Nextstrain command.

`sh` is probably actually `zsh` or `bash` in POSIX-compat mode. We clearly need to be using a different file, but I'm not sure if `~/.profile` is the right one as I suspect that's only used on login shells (e.g. as it is by `bash`, at least in non-POSIX-compat mode). On macOS, all shells in Terminal.app are login shells by default, but that's not true for elsewhere.

First step is to reproduce this locally on macOS and Linux. Next is to figure out the most appropriate file.

corneliusroemer:
Off the top off my head, thinking about why someone would get `zsh` in POSIX-compat mode: maybe if one doesn't have `xcode-select` tools installed things differ from the average Nextstrain dev mac setup.

Is there a reason we currently pipe into `bash` on macOS? `zsh` would be more canonical on mac as the default bash is now very outdated.

```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.

$ /bin/bash
The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ 
```

corneliusroemer:
[...]

Issue #320, opened 2023-10-10
Broken CI due to upstream release; mypy implicated


tsibley:
[This morning's scheduled CI run](https://github.com/nextstrain/cli/actions/runs/6472933808) failed across Linux/macOS/Windows on the "test-source" jobs. The errors are limited to `tests/mypy.py`, which runs mypy on the codebase.

Given the errors are about things that used to work just fine, this is likely due to a new upstream release which changed behaviour. Probably mypy itself by the looks of it (but potentially `importlib_metadata` instead/too).

tsibley:
Indeed, Mypy 1.6.0 was released this morning. Two things stand out in [the changelog](https://mypy-lang.blogspot.com/2023/10/mypy-16-released.html) relative to the errors we see:

> Running mypy with `--python-version 3.6`, for example, is no longer supported. Python 3.6 hasn’t been properly supported by mypy for some time now, and this makes it explicit.

and

> Mypy now does a better job inferring type variables inside arguments of callable types. 

One solution is to pin to `mypy <1.6` for now.

Another is to ditch Mypy and stick with just Pyright for now. When I adopted Mypy and then Pyright a short time later, it was because both were newer and not entirely overlapping in functionality. There was extra checking we got from running both. At this point though, Pyright is (IMO) superior for our use, and mostly we work around Mypy issues.

Separately, it's likely also time to ditch Python 3.6. See https://github.com/nextstrain/cli/issues/136#issuecomment-1622294693.

PR #321, opened 2023-10-10
standalone-installer-unix: Use ~/.profile when the default shell is sh

tsibley:
We were suggesting nonsense (~/.shrc) due to our final fallback/guess. Instead, handle sh specifically and correctly.

Resolves <https://github.com/nextstrain/cli/issues/319>.

## Checklist

- [x] Manual test with default shell of `/bin/sh` works
- [x] Checks pass


tsibley:
CI will probably fail due to #320

victorlin (approved):

PR #323, opened 2023-10-11
authn: Handle case when a user is not in any group

victorlin:

In this case, the "cognito:groups" claim does not exist. Previously, this resulted in a KeyError upon login and whoami commands.

## Related issue(s)

N/A

## Checklist

- [x] `nextstrain login` and `nextstrain whoami` works locally for group-less user in testing environment
- [ ] Checks pass

### DIFF
--- a/nextstrain/cli/authn.py
+++ b/nextstrain/cli/authn.py
@@ -48,7 +48,7 @@ class User:
         assert session.id_claims
 
         self.username = session.id_claims["cognito:username"]
-        self.groups   = session.id_claims["cognito:groups"]
+        self.groups   = session.id_claims.get("cognito:groups", [])
         self.email    = session.id_claims["email"]
 
         self.http_authorization = f"Bearer {session.id_token}"


nextstrain/cli/authn.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, session: cognito.Session):
         assert session.id_claims
 
         self.username = session.id_claims["cognito:username"]
-        self.groups   = session.id_claims["cognito:groups"]
+        self.groups   = session.id_claims.get("cognito:groups", "")
```
tsibley, line 51:
```suggestion
        self.groups   = session.id_claims.get("cognito:groups", [])
```

The type of `self.groups` is `List[str]`

https://github.com/nextstrain/cli/blob/1247a9be07939bcbf59394371112af1f1706adf4/nextstrain/cli/authn.py#L38-L45

as the `cognito:groups` claim is a JSON array of strings.

tsibley (changes requested):
Nice little bug fix! Thanks. One minor tweak to the fallback value.

nextstrain/cli/authn.py
tsibley: Not sure why the type checking didn't catch that, but it's not entirely surprising as the type checking isn't comprehensive, just "best effort".

victorlin: Thanks, good catch! [Fixed](https://github.com/nextstrain/cli/compare/1ccbed2bf6624e41cc2399920fbd789cf856dfd4..9cbb8042874f60dafe812527d923d13664782b21).

I think it's because `session.id_claims` has an implicit type of `Any`, so type checking is [ignored](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code) on that entire expression.

PR #256, opened 2023-02-24
Bump minimum Python version from 3.6 to 3.7

corneliusroemer:
Mirrors previous bump commit https://github.com/nextstrain/cli/commit/d7d819674269385537095f23a9591d44fdfedd98

Motivated by failing CI for a PR with 3.6 and by the fact that 3.6's end of life was 23 Dec 2021

A few places in code which handle 3.6 vs. 3.7+ differences still
remain, as they're not enough overhead to rip out (yet).

This change warrants a major version bump for the first release
including it.

PR #325, opened 2023-11-03
Bump minimum Python version to 3.8

### LINKED ISSUE #136: Bump minimum Python dependency to 3.7 (or newer)
## Context

We currently require Python 3.6 or higher, but 3.6 will reach end of life in three weeks.

## Description

We should require at least version 3.7 now.

## Possible solution

 - [ ] Update setup.py to set minimum Python version to 3.7
 - [ ] Update GitHub CI workflow to drop 3.6
 - [ ] Update change log noting breaking change
 - [ ] Release a new major version of CLI to reflect breaking change
 - [ ] [Create a PR](https://github.com/nextstrain/bioconda-recipes/) to update [the minimum Python version in the CLI's bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/fca9554c95763983bdc9fb06770f956b90967057/recipes/nextstrain-cli/meta.yaml#L20-L24)

Making the change to the bioconda repository first could potentially cause issues for users, although releasing a new CLI version will trigger an autobump of the bioconda package that we will have to supersede with our own PR.

victorlin:

Drop support for Python 3.6 and 3.7. Adding support for newer versions will be done in separate PRs.

## Related issue(s)

- Closes #136
- Previous attempt at #256

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,19 +35,11 @@ jobs:
           - macos-latest
           - windows-latest
         python:
-          - '3.6'
           - '3.7'
           - '3.8'
           - '3.9'
           - '3.10'
 
-        # Python 3.6 is not available on Ubuntu 22.04, which is what
-        # ubuntu-latest points to as of Q4 2022¹, so replace that job with 3.6
-        # on Ubuntu 20.04.
-        #   -trs, 22 Nov 2022
-        #
-        # ¹ <https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/>
-        #
         # A bodged https://github.com/actions/python-version build of 3.7.17
         # for macOS mistakenly doesn't include the bz2 module of the stdlib¹,
         # which breaks us because some of our deps (fsspec, at the least)
@@ -59,13 +51,9 @@ jobs:
         # ² <https://github.com/nextstrain/cli/pull/289#issuecomment-1595417903>
         # ³ <https://github.com/fsspec/filesystem_spec/pull/1295>
         exclude:
-          - os: ubuntu-latest
-            python: '3.6'
           - os: macos-latest
             python: '3.7'
         include:
-          - os: ubuntu-20.04
-            python: '3.6'
           - os: macos-latest
             python: '3.7.16'
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


tsibley:
Would you add an entry to the changelog for this PR?

nextstrain/cli/__main__.py
```diff
@@ -13,30 +13,24 @@ def main():
...
 
     Suitable only for output streams (e.g. stdout, stderr), as reconfiguring an
     input stream is more complicated.
     """
-    # XXX TODO: When we drop Python 3.6 support, most of this function can be
-    # replaced by stdio.reconfigure().
-    #   -trs, 6 June 2022
-
     # Flush any pending output under old configuration.
     stdio.flush()
```
tsibley, line 30: This can go. From the `.reconfigure()`'s methods docs:

> This method does an implicit stream flush before setting the new parameters.

Similarly, the explicit setting of `line_buffering` below can go away, as it will be preserved by using `.reconfigure()`.

.github/workflows/ci.yaml
```diff
@@ -213,7 +213,7 @@ jobs:
         python:
           - '3.8'
           - '3.9'
-          # XXX TODO: Add 3.10 here once supported by Conda/Bioconda/Conda Forge.
+          - '3.10'
```
tsibley, line 216: The commit message for this change is titled:

>  CI: Enable Python 3.10 testing of Conda package 

But there's no testing of a Conda package here. The reason Conda is implicated in the removed comment is because the `./src/.github/actions/setup-integration-tests` action used in this job relies on Conda to set up the test environment. But the actual package the job tests is a Python package, not a Conda package.

tsibley (approved):
Looks great! A couple minor things noted. A changelog entry as part of this PR would be great too. When released, this will warrant a major version bump, naturally.

nextstrain/cli/__main__.py
victorlin: Thanks for the tips, updated in 58104afa0c253fad48185437ae93c54210085288.

.github/workflows/ci.yaml
victorlin: Oh, I remember being confused - that makes a lot more sense now. Updated the message in ea0b76674d4ce0bc30201b18d9928283a67e0139

victorlin:
Changelog entry added in e15afcca46ac22eb4951f0515160700875887d13.

PR #326, opened 2023-11-06
Cast err.response to Response class for Pyright

victorlin:

2 weeks ago¹, Pyright on 3.10 CI started failing with an error:

    "status_code" is not a known member of "None"

I looked at the requests source code² and it makes sense. However, I was not able to reproduce the CI error locally and don't understand why it only happens on 3.10. Hopefully this addresses the issue.

¹ https://github.com/nextstrain/cli/actions/runs/6605300519/job/17940321340#step:5:88
² https://github.com/psf/requests/blob/839a8edec37c81a18ac8332cfbd44f44e1ae6206/src/requests/exceptions.py#L19

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/remote/nextstrain_dot_org.py
+++ b/nextstrain/cli/remote/nextstrain_dot_org.py
@@ -683,6 +683,7 @@ def raise_for_status(response: requests.Response) -> None:
         response.raise_for_status()
 
     except requests.exceptions.HTTPError as err:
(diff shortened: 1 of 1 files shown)


tsibley (approved):
Thanks! I'd been meaning to get around to looking at the CI failures but hadn't been forced to do so yet by other work. It'll be nice for it to work again!

PR #330, opened 2023-11-16
remote: Add zstd to encoding types

joverlee521:
Allows us to skip gzip compression when uploading .zst files. Includes adding ".zst" to the mimetypes encoding types dictionary because it is not yet included in the standard library.¹

¹ https://github.com/python/cpython/blob/2ef3676a5bb8fba531fb8237ce50c27ebe37fb96/Lib/mimetypes.py#L411-L417

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/remote/s3.py
+++ b/nextstrain/cli/remote/s3.py
@@ -65,6 +65,9 @@ from ..types import S3Bucket, S3Object
 mimetypes.add_type("application/json", ".json")
 mimetypes.add_type("text/markdown", ".md")
 
+# Add zstd to encodings map since it is not yet included in the standard library
+mimetypes.encodings_map[".zst"] = "zstd"
+
 
 def upload(url: urllib.parse.ParseResult, local_files: List[Path], dry_run: bool = False) -> Iterable[Tuple[Path, str]]:
     """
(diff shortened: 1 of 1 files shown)


joverlee521:
Tested locally with 
```
nextstrain remote upload s3://nextstrain-data/files/dengue/trial/ /tmp/sequences_all.fasta.zst
```

Seems to have worked as expected
```
$ curl -I https://data.nextstrain.org/files/dengue/trial/sequences_all.fasta.zst
HTTP/2 200 
content-type: application/zstd
```

nextstrain/cli/remote/s3.py
```diff
@@ -65,6 +65,9 @@
 mimetypes.add_type("application/json", ".json")
 mimetypes.add_type("text/markdown", ".md")
 
+# Add zstd to encodings map since it is not yet included in the standard library
+mimetypes.encodings_map[".zst"] = "zstd"
```
tsibley, line 69: _non-blocking, just a comment_

This addition is just fine, but seeing it and the existing additions above it reminds me that this module should probably be instantiating its own [`MimeTypes` instance](https://docs.python.org/3/library/mimetypes.html#mimetypes-objects) and modifying that rather than modifying the shared process-global instance.

tsibley (approved):
LGTM! Thanks for adding this.

nextstrain/cli/remote/s3.py
joverlee521: Tracking in https://github.com/nextstrain/cli/issues/331.

PR #332, opened 2023-11-21
Various and sundry

tsibley:
See commit messages. Done during work on the path towards OIDC/OAuth2-based login, but suitable for separate review and independent merge.

## Checklist

- [x] Checks pass

### DIFF
--- a/devel/generate-command-doc
+++ b/devel/generate-command-doc
@@ -46,6 +46,10 @@ os.environ.update({
 
     # Ensure we detect a browser for stable `nextstrain view` output.
     "BROWSER": "/bin/true",
+
+    # Ensure HOST and PORT are stable for `nextstrain view` output.
+    "HOST": "127.0.0.1",
+    "PORT": "4000",
 })
 
 from nextstrain.cli import make_parser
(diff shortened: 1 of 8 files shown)


tsibley:
I was going to write up change log entries, but all of these seem like internal improvements to me and not user-facing. The one or two that might be user-facing are only so when things are used in development contexts (e.g. with `NEXTSTRAIN_DOT_ORG` set).

victorlin (approved):
Local testing:

- [x] `nextstrain view` works with 3261aacdd8a1c564c4db392d7388eab871d2349e
- [x] `NOBROWSER=1 nextstrain view` works with 32f06feca1608c3f90af17e1c28883e63243d845

joverlee521 (approved):
I'm probably missing context from the ongoing CDC work, but changes LGTM from inspection. 

I'm curious what prompted the `NOBROWSER` variable?

tsibley:
> I'm curious what prompted the `NOBROWSER` variable?

When doing browser work for #333 and `nextstrain login`, I noticed that setting `BROWSER=` didn't work as I expected it to. It doesn't result in finding no browser and thus not automatically opening one, instead it's just an empty list of preferred browsers and so still falls back to the built-in defaults.

Since `BROWSER` is handled by the `webbrowser` module in the stdlib, I added `NOBROWSER`. This is an env var that's somewhat of a convention, akin to [`NOCOLOR`](https://nocolor.org) (though not as widely supported).

PR #335, opened 2023-12-06
Address new Pyright error

victorlin:

The latest version of Pyright exposed a new error for this project:

    argparse.py:52:9 - error: Method "_fill_text" overrides class "HelpFormatter" in an incompatible manner
      Parameter 4 name mismatch: base parameter is named "indent", override parameter is named "prefix" (reportIncompatibleMethodOverride)

I don't think the issue resulted in badly rendered docs, but nevertheless it's easy to address this by changing a couple names.

## Related issue(s)

Fixes the issue that caused [today's scheduled CI](https://github.com/nextstrain/cli/actions/runs/7118237967) to fail.

## Checklist

- [x] Checks pass


victorlin:
Merging to have successful CI.

PR #341, opened 2023-12-27
remote: Show more possible reasons for S3 access error

### LINKED ISSUE #56: [deploy] Hint that a bucket may not be found because of missing/wrong credentials
when running deploy without or wrong credentials, the error message is not "not authenticated" but "bucket doesn't exist"...
```
\:> nextstrain deploy s3://nextstrain-staging auspice/v2_flu_seasonal_*
No bucket exists with the name "nextstrain-staging".

Buckets are not automatically created for safety reasons.
```
not sure whether there is a way around this, but not the ideal error message...

victorlin:

Previously, other reasons detailed in the changes would show an error message saying "no bucket exists" even if it does exist.

Update the error text to describe more possible reasons based on docs for S3.Client.head_bucket.¹ and local testing. This list might not be exhaustive, but I can't come up with any other scenarios.

¹ <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_bucket.html>

## Related issue(s)

- [Internal inquiry](https://bedfordlab.slack.com/archives/C036L7V615F/p1703710506208669)
- Closes #56

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/remote/s3.py
+++ b/nextstrain/cli/remote/s3.py
@@ -225,9 +225,11 @@ def split_url(url: urllib.parse.ParseResult) -> Tuple[S3Bucket, str]:
 
     except ClientError:
         raise UserError(f"""
-            No bucket exists with the name "{bucket.name}".
+            Unable to read from S3 bucket "{bucket.name}". Possible reasons:
 
-            Buckets are not automatically created for safety reasons.
+            1. Your AWS credentials are invalid.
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
LGTM! Too bad the `ClientError` does return the different error codes.

jameshadfield (approved):

jameshadfield:
This PR will close #56

Issue #329, opened 2023-11-09
Singularity/Apptainer Version


osageorange:
The Singularity version check in the referenced line prevents using the Apptainer flavor of Singularity because the version numbers were restarted after the fork occurred in 2021 -- the current version of Apptainer is 1.2, hence it fails the check. A comment on lines 34 - 39 suggest that either flavor will work. Can we modify the version check to enable Apptainer usage?

Thanks!

https://github.com/nextstrain/cli/blob/4dc06ebd7db130ed521550e78342f6a2f18dc7a2/nextstrain/cli/runner/singularity.py#L120C38-L120C38

tsibley:
Thanks for flagging this. Indeed, we intend to try to maintain compatibility with both SingularityCE and Apptainer. I didn't realize/recall that Apptainer restarted numbering when I added Singularity version inspection!

It looks like Apptainer 1.0.0 surpasses our minimum version since it's based on Singularity 3.9.5.

It's not just that base minimum version check we'll need to adapt, however, as we also check the version for optional features we conditionally enable. (Grep for `singularity_version_at_least` to see what I mean.)

tsibley:
Something like this patch might work, but I haven't tested it.

```diff
diff --git a/nextstrain/cli/runner/singularity.py b/nextstrain/cli/runner/singularity.py
index 535fa36..53d1f6d 100644
--- a/nextstrain/cli/runner/singularity.py
+++ b/nextstrain/cli/runner/singularity.py
@@ -119,6 +119,8 @@ DEFAULT_IMAGE = os.environ.get("NEXTSTRAIN_SINGULARITY_IMAGE") \
 
 SINGULARITY_MINIMUM_VERSION = "3.0.0"
 
+APPTAINER_MINIMUM_VERSION = "1.0.0"
+
 SINGULARITY_CONFIG_ENV = {
     # Store image caches in our runtime root instead of ~/.singularity/…
     "SINGULARITY_CACHEDIR": str(CACHE),
@@ -158,7 +160,7 @@ def SINGULARITY_EXEC_ARGS(): return [
     #     --writable-tmpfs        (3.0.0)
     #     --no-init               (3.0.0)
     #     --no-umask              (3.7.0)
-    #     --no-eval               (3.10.0)
+    #     --no-eval               (3.10.0, Apptainer 1.1.0)
     #
     # We opt not to use the --compat bundle option itself mainly for broader
     # version compatibility but also because what it includes will likely
[...]

PR #343, opened 2024-01-10
runner.singularity: Explicitly support Apptainer's `singularity` when checking versions

tsibley:
My intent was (and remains) to try to maintain compatibility with both SingularityCE and Apptainer after the Singularity fork.  I didn't realize/recall that Apptainer restarted version numbering when I added this version inspection!

It would be most ideal if Apptainer's `singularity` executable reported the equivalent/compatible SingularityCE version, but that's understandably difficult and so I get why it reports the Apptainer version instead.

Thanks to @osageorange for raising this issue, testing this patch, and reporting back that it works!

As SingularityCE and Apptainer continue to diverge, a more complicated treatment of them may be needed, but for now this should suffice.

Resolves <https://github.com/nextstrain/cli/issues/329>.

## Checklist

- [x] Checks pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,13 @@ This release drops support for Python versions 3.6 and 3.7 and adds support for
   when the default shell is sh.
   ([#321](https://github.com/nextstrain/cli/pull/321))
 
+* The Singularity runtime once again supports Apptainer's `singularity`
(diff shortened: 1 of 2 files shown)


tsibley:
CI was failing type checking for changes in this PR (now fixed) and also for unrelated reasons (fixed in #344).

victorlin (approved):

PR #346, opened 2024-01-18
devel/create-github-release: Reflow paragraphs in changelog Markdown to single long lines

tsibley:
This is necessary to work around GitHub's rendering of release notes which treats newlines in the Markdown as hard line breaks instead of soft line breaks (which is more standard).  GitHub itself isn't consistent and treats newlines as hard or soft in different places.¹

This has bugged me for a while, and I've always resorted to hand-correcting the release notes after the release is created.  But that's annoying and means that the crappy formatting still appears in emails and other notifications triggered by the release creation event. I'll be glad to have Pandoc doing this for me instead!  It hadn't occurred to me to use Pandoc for this until now.

¹ <https://github.com/orgs/community/discussions/57633>

## Checklist

- [x] Checks pass

### DIFF
--- a/devel/create-github-release
+++ b/devel/create-github-release
@@ -31,7 +31,7 @@ main() {
         --title "$version" \
         --target "$commit" \
         ${pre_release:+--prerelease} \
-        --notes-file <(preamble; "$devel"/changes "$version") \
+        --notes-file <(preamble; "$devel"/changes "$version" | "$devel"/reflow-markdown) \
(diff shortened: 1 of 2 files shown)


victorlin (approved):

tsibley:
[8.0.1](https://github.com/nextstrain/cli/releases/tag/8.0.1) is the first release to use this, and it appears to have worked a treat.

PR #353, opened 2024-01-31
Mention Singularity in a few more lists of runtimes that occur in help and docs

tsibley:
Good to be consistent when enumerating our runtimes, even if a couple of the lists in context suggest they're non-exhaustive.

## Checklist

- [x] Checks pass

### DIFF
--- a/doc/commands/build.rst
+++ b/doc/commands/build.rst
@@ -22,8 +22,8 @@ The build directory should contain a Snakefile, which will be run with
 snakemake.
 
 You need at least one runtime available to run a build.  You can test if the
-Docker, Conda, ambient, or AWS Batch runtimes are properly supported on your
-computer by running::
+Docker, Conda, Singularity, ambient, or AWS Batch runtimes are properly
+supported on your computer by running::
 
     nextstrain check-setup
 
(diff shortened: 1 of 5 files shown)


victorlin:
Is it also worth mentioning on the main docs [install page](https://docs.nextstrain.org/en/latest/install.html)? Example:

image

tsibley:
Probably!

j23414:
[non-blocking comment]

Despite the [rebranding of "Singularity" to "Apptainer,"](https://apptainer.org/) I assume that the majority of users are still more acquainted with the name "Singularity."

victorlin:
@j23414 If you have access to any HPC systems, do you know if `singularity` or `apptainer` (or both) are available as commands?

j23414:
I have access to 2 HPCs, so some but limited experience.  

For one, both are available where 
> the `singularity` command is just a symlink to `apptainer`

The other seems to maintain a separate `apptainer` module: https://sciwiki.fredhutch.org/compdemos/Apptainer/

Notably, my previous pipelines dependent on Singularity are now generating various migration warnings, such as advising the use of 'APPTAINER_CACHEDIR' instead of 'SINGULARITY_CACHEDIR.' which people msg/email me about 😆

j23414:
In summary, I think it's safe to use the term "singularity" or "Apptainer (was singularity)".

tsibley:
Yeah, the [Singularity runtime documentation](https://docs.nextstrain.org/projects/cli/en/stable/runtimes/singularity/?highlight=Apptainer) page discusses Singularity vs. Apptainer and compatibility.

PR #355, opened 2024-01-31
Better support pathogen repositories which place workflows in subdirectories

tsibley:
…as this is the direction we're moving.

The main change is making the filesystem isolation boundary (i.e. what's mapped to `/nextstrain/build` in a container) a _separate thing_ from the workflow and initial working directory (i.e. what's given to `nextstrain build`).  In this codebase, these two things are referred to as the _build volume_ (aka `opts.build`) and the _working volume_. Historically, the working volume given to the runners for `nextstrain build` and `nextstrain shell` _was_ the build volume; now, they're separately considered and sometimes differ.

See the included changelog entry for usage details and rationale.  For background, I made the initial proposal for this feature¹ on a PR in our pathogen-repo-template repository and some discussion ensued.

¹ <https://github.com/nextstrain/pathogen-repo-template/pull/16#issuecomment-1911079700>
  or <https://github.com/tsibley/blab-standup/blob/master/2024-01-25.md>

## Related issue(s)

Based on #354.

## Checklist

- [x] Checks pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,67 @@ development source code and as such may not be routinely kept up to date.
 
 # __NEXT__
 
+## Features
+
+* `nextstrain build` and `nextstrain shell` now better support pathogen
+  repositories which place workflows in subdirectories.  The top-level of the
+  repo must contain a `nextstrain-pathogen.yaml` file for this support to
+  activate.  The file may be empty for now, though we anticipate using it for
+  pathogen-level metadata in the future to aid indexing, listing, and
+  attribution of pathogen repos.
+
+  As an example of the new support, consider the following repo layout
+
+      mpox/
+      ├── nextstrain-pathogen.yaml
+      ├── ingest/
+      │   ├── Snakefile
+      │   └── …
+      ├── phylogenetic/
+      │   ├── Snakefile
+      │   └── …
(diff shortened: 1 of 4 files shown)


tsibley:
Some new tests are failing on Windows, due to path handling.

```
Expected:
    NamedVolume(name='build/ingest', src=...Path('.../tests/data/pathogen-repo/ingest'), dir=True, writable=True)
Got:
    NamedVolume(name='build\\ingest', src=WindowsPath('D:/a/cli/cli/tests/data/pathogen-repo/ingest'), dir=True, writable=True)
```

I'll fix it up.

huddlej (approved):
I don't have strong feelings about the technical implementation, but I love the idea of the top-level indicator file that works akin to `.git` to designate a pathogen directory. I also love that this approach allows us to run the Nextstrain CLI the way that feels most natural (to me, at least) by specifying the directory with the workflow we want to run (`nextstrain build ingest`) and not needing to also specify the path to the Snakefile (`nextstrain build . -s ingest/Snakefile` which reveals implementation details that we don't necessarily want to formalize).

I also love that you made this happen so quickly after our random coffee/lab chat about it, @tsibley, even if it was to head-off my terrible suggestions involving symlinks! 🌈

PR #356, opened 2024-01-31
volume: store_volume("X") no longer sets opts.X

tsibley:
The volume named "X" is now only appended to the `opts.volumes` list.

The last usage of the prior functionality, for `opts.build`, was removed in "Better support pathogen repositories which place workflows in subdirectories" (bf858d2).

I think this clarifies a potentially-confusing part of the argument parsing too (e.g. "where did `opts.X` come from?").

## Related issue(s)

Based on #355.

## Checklist

- [ ] Checks pass

### DIFF
--- a/nextstrain/cli/volume.py
+++ b/nextstrain/cli/volume.py
@@ -23,8 +23,7 @@ def store_volume(volume_name):
     path definitions for named volumes.
 
     Each named volume is stored as a NamedTuple (name, src).  The tuple is
-    stored on the options object under the volume's name (modified to replace
-    slashes with underscores), as well as added to a shared list of volumes,
+    stored on the options object as an element in a shared list of volumes,
     accessible via the "volumes" attribute on the options object.
 
     For convenient path manipulation and testing, the "src" value is stored as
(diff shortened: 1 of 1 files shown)


nextstrain/cli/volume.py
```diff
@@ -37,7 +36,4 @@ def __call__(self, parser, namespace, values, option_strings = None):
             new_volume = NamedVolume(volume_name, Path(values)) if values else None
             setattr(namespace, "volumes", [*volumes, new_volume])
 
-            # Allow the new volume to be found by name on the opts object
-            setattr(namespace, volume_name.replace('/', '_'), new_volume)
```
victorlin, line 41: Curious - what was the purpose of `replace('/', '_')`? I see that `volume_name` takes on values `augur`, `auspice`, `fauna`, `sacra`, and previously `build` - none of which were affected by this replacement.

victorlin (approved):

tsibley:
@victorlin Thanks for your comment. I'd totally forgotten about this minor change and hadn't swept the PR list recently for this repo.

nextstrain/cli/volume.py
tsibley: Prior to a0f3ad197f0d3946c090ed51b0230ea30ac68776 there [was](https://github.com/nextstrain/cli/commit/a0f3ad197f0d3946c090ed51b0230ea30ac68776#diff-8846c03699df577b0083c116e35a2dcde0f3093f6e58606f9584de266f77cdc7L126-R138) a `store_volume("auspice/data")` call for `nextstrain view`.

PR #357, opened 2024-02-01
Standalone builds for macOS aarch64 (aka arm64, Apple Silicon, M1/M2)

tsibley:
GitHub released public runners.¹  Great timing too, as I was just about
to look into using the ARM hardware that's available for macOS with
GitHub's (paid) large runners².  This also beats operating a self-hosted
runner, whether that's on EC2³ or our own Mac Mini.⁴

And the icing on the cake is that everything Just Worked by merely
enabling builds for this architecture.  \o/  High fives all around,
everyone.

Changes to the standalone installer to follow separately⁵, as those need
to be held for merge until after the first release with these builds
available.

¹ <https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/>
² <https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1706556690621389>
³ <https://github.com/tsibley/blab-standup/blob/e7e2516fc40856a94b54eb83077d8f4bafd16914/2022-09-14.md>
⁴ <https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1670979565748349?thread_ts=1670963819.691499&cid=C01LCTT7JNN>
⁵ <https://github.com/nextstrain/cli/pull/358>

## Checklist

- [x] Checks pass

PR #361, opened 2024-02-02
Check if Rosetta 2 is enabled in the Conda and Docker runner checks

tsibley:
When on macOS on aarch64 hardware, the Conda runtime currently requires Rosetta.  The Docker runtime does not require it, but can benefit from it because our aarch64 image isn't totally free of x86_64 code (yet?), so emulation is required and Rosetta is faster than QEMU.

The standalone installer still checks for Rosetta as we previously didn't produce aarch64 standalone binaries¹, but that check will be made conditional on Nextstrain CLI version in a subsequent commit² after the first release of aarch64 support and these new Rosetta checks.

¹ <https://github.com/nextstrain/cli/pull/357>
² <https://github.com/nextstrain/cli/pull/358>

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/runner/docker.py
+++ b/nextstrain/cli/runner/docker.py
@@ -87,7 +87,7 @@ from typing import Iterable, List
 from .. import config, env
 from ..errors import UserError
 from ..types import Env, RunnerSetupStatus, RunnerTestResults, RunnerTestResultStatus, RunnerUpdateStatus
-from ..util import warn, colored, capture_output, exec_or_return, split_image_name
+from ..util import warn, colored, capture_output, exec_or_return, split_image_name, test_rosetta_enabled
 from ..volume import store_volume, NamedVolume
 from ..__version__ import __version__
 
@@ -387,7 +387,12 @@ def test_setup() -> RunnerTestResults:
         ('docker run works',
             test_run()),
         *test_memory_limit(),
-        *test_image_version()
+        *test_image_version(),
+
+        # Rosetta 2 is optional, so convert False (fail) → None (warning)
+        *[(msg, None if status is False else status)
+            for msg, status
+             in test_rosetta_enabled("Rosetta 2 is enabled for faster execution (optional)")],
     ]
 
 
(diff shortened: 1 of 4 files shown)


tsibley:
I've tested this on the Blab Mac mini running macOS 12 on aarch64. It works as expected there. Going to merge, but I'd welcome post-merge review!

nextstrain/cli/runner/docker.py
```diff
@@ -387,7 +387,12 @@ def test_image_version():
         ('docker run works',
             test_run()),
         *test_memory_limit(),
-        *test_image_version()
+        *test_image_version(),
+
+        # Rosetta 2 is optional, so convert False (fail) → None (warning)
+        *[(msg, None if status is False else status)
+            for msg, status
+             in test_rosetta_enabled("Rosetta 2 is enabled for faster execution (optional)")],
```
victorlin, line 395: At first I wondered, why not return a single `RunnerTestResult` instead of list from `test_rosetta_enabled()`? Then I realized that it allows same-line list comprehension conversion like you've done here. But what about for `*test_memory_limit()` which could be simply `test_memory_limit()`?

```suggestion

```

victorlin (commented):
- [x] Tested using local git installation

```
$ nextstrain check-setup
nextstrain-cli is up to date!

Testing your setup…

# docker is supported
✔ yes: docker is installed
✔ yes: docker run works
✔ yes: containers have access to >2 GiB of memory (limit is 7.8 GiB)
✔ yes: image is new enough for this CLI version
✔ yes: Rosetta 2 is enabled for faster execution (optional)

# conda is supported
✔ yes: operating system is supported
✔ yes: Rosetta 2 is enabled
✔ yes: runtime data dir doesn't have spaces
✔ yes: runtime appears set up
✔ yes: snakemake is installed and runnable
✔ yes: augur is installed and runnable
✔ yes: auspice is installed and runnable
```

nextstrain/cli/runner/docker.py
tsibley: I think `*test_memory_limit()` (and `*test_image_version()`) may be historic holdovers from a previous iteration of the code here where those functions sometimes returned `[(msg, status)]` but other times returned `[]`. I haven't dug into history though…

tsibley: Note that `test_rosetta_enabled()` returns a list of results because it doesn't always want to include a Rosetta test on all platforms:

https://github.com/nextstrain/cli/blob/25075faebad904ff3d9df9422abb16bedb662f80/nextstrain/cli/util.py#L611-L612

PR #365, opened 2024-04-25
runner: Support NEXTSTRAIN_RUNTIME_ENVDIRS as an alternative to --envdir

tsibley:
Motivated by some of our automation's desire to pass a set of env vars into the runtime without having to munge a user-provided `nextstrain build` invocation.¹

¹ <https://github.com/nextstrain/private/issues/96#issuecomment-2046243636>

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/runner/__init__.py
+++ b/nextstrain/cli/runner/__init__.py
@@ -1,5 +1,6 @@
 import argparse
-from argparse import ArgumentParser
+import os
+from argparse import ArgumentParser, ArgumentTypeError
 from typing import cast, List, Union, TYPE_CHECKING
 # TODO: Use typing.TypeAlias once Python 3.10 is the minimum supported version.
 from typing_extensions import TypeAlias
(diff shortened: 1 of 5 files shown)


nextstrain/cli/runner/__init__.py
```diff
@@ -264,6 +266,14 @@ def run(opts: Options, working_volume: NamedVolume = None, extra_env: Env = {},
     if opts.__runner__ is singularity and opts.image is docker.DEFAULT_IMAGE: # type: ignore
         opts.image = singularity.DEFAULT_IMAGE # type: ignore
 
+    if envdirs := os.environ.get("NEXTSTRAIN_RUNTIME_ENVDIRS"):
+        try:
+            opts.envdir = [
+                *[DirectoryPath(d) for d in envdirs.split(os.pathsep) if d],
+                *opts.envdir ]
```
victorlin, line 273: Is there a reason to prepend here and not `opts.envdir.extend`?

victorlin (approved):
Skimmed changes and didn't test, looks good

nextstrain/cli/runner/__init__.py
tsibley: Yep. It follows the traditional Unix cascade of: built-in default → environment variable → command line option. So for example, if you provide an env var `A` via `NEXTSTRAIN_RUNTIME_ENVDIRS` and also again via `--envdir`, the value from the latter will be used.

tsibley:
@joverlee521 This will be part of 8.3.0 once CI completes.

Issue #219, opened 2022-09-16
[batch] exclude directories for upload


jameshadfield:
### Context

Deploying a build to AWS-batch currently zips up (almost) the entire current directory and uploads this to S3. [Certain assets are not uploaded, e.g. `.git/`, but these are not user-definable](https://github.com/nextstrain/cli/blob/master/nextstrain/cli/runner/aws_batch/s3.py#L50-L65). I commonly have lots of large files which are unnecessary for the current build and it would be quicker and less overhead to ignore them for the purposes of the desired build. As examples: `./data` (sometimes >100Gb), big JSONs in `./auspice` (can take more than an hour to zip), and a _huge_ number of intermediate files in `./results`.

This causes pain during job deployment (both time to zip up, and time to upload the zip file) as well as during download (as the zip file is bigger than necessary). 

My current workflow often involves creating a "temporary" directory which is a copy of the the current one without those directories, or moving the directories to a temporary place while the job is deployed. Both are a pain and prone to messing up badly!

### Description
Add a `--exclude-from-upload` flag, which I'd commonly use use like so: `--exclude-from-upload auspice results data`

I'm not sure how this would work when downloading if you had a previous `auspice/` directory which wasn't part of the uploaded assets, and then downloaded a (aws-batch generated) `auspice/` directory. I'm not that familiar with the logic for working out which files to download (I typically request a subset of files to be downloaded).

corneliusroemer:
I agree that we don't need to update things that are just downloaded - like `data`, but results are often useful for debugging or quickly rerunning a rule locally without having to rerun time-intensive steps.

Also, we may reduce storage by `tar.zst` compressing instead of using standard `zip`. Since we have lots of sequences `zstd` compression would work extremely well.

jameshadfield:
> but results are often useful for debugging or quickly rerunning a rule locally without having to rerun time-intensive steps

Yup, I agree, but I think we're thinking of slightly different things. This is a request to avoid _uploading_ certain artefacts via an opt-in process -- e.g. in my case `./results` from previous runs which aren't relevant to the current job I want to deploy.

tsibley:
I think the confusion may be that @jameshadfield is talking about the upload from the local computer → S3 before the Batch job is submitted and @corneliusroemer is talking about the upload from the remote Batch job → S3 at workflow completion.

corneliusroemer:
[...]

PR #370, opened 2024-05-24
build: Add `--exclude-from-upload` and `--exclude-from-download` options

tsibley:
Excluding files from upload is very handy for ad-hoc skipping of large ancillary files or previous output files in the build directory that the user wants to ignore for the remote build on AWS Batch (i.e. to start it "fresh").  Existing workarounds for the lack of an exclusion mechanism include git worktrees to obtain a clean state and moving files or directories temporarily out of the build directory.

A future improvement would be adding support to also specify these patterns via a file, which can then be checked in to a build repo and shared by all users.

As a broader improvement, I'd like to design away this issue by separating out the workflow-as-program (rules, code, etc.) from the workflow-as-state (config, inputs, outputs, etc.).  This comes out of several other goals, but would apply to this "excluded files" need too. For example, instead of relying on the implicit state combined with the code like now in a single build directory, we'd instead explicitly pass an empty starting state separate from the workflow program.

This change includes a small bug fix for --download to allow _only_ negated patterns.

Resolves: <https://github.com/nextstrain/cli/issues/219>

## Checklist

- [x] Checks pass

### DIFF
--- a/doc/commands/build.rst
+++ b/doc/commands/build.rst
@@ -97,12 +97,66 @@ options
     wildcards (``**``), extended globbing (``@(…)``, ``+(…)``, etc.),
     and negation (``!…``).
 
+    Patterns should be relative to the build directory.
+
 
 
 
 .. option:: --no-download
 
     Do not download any files from the remote build when it completes. Currently only supported when also using :option:`--aws-batch`.
+
+.. option:: --exclude-from-download <pattern>
+
+    Exclude files matching ``<pattern>`` from being downloaded from
+    the remote build.  Equivalent to passing a negated pattern to
+    :option:`--download`.  That is, the following are equivalent::
+
+        --exclude-from-download 'xyz'
+        --download '!xyz'
+
+    Refer to :option:`--download` for usage details, but note that this
+    option doesn't support already-negated patterns (e.g. ``!…`` or
+    ``!(…)``).
+
+    This option exists to parallel :option:`--exclude-from-upload`.
+
+
+
+
+.. option:: --exclude-from-upload <pattern>
+
+    Exclude files matching ``<pattern>`` from being uploaded as part of
+    the remote build.  Shell-style advanced globbing is supported, but
+    be sure to escape wildcards or quote the whole pattern so your
+    shell doesn't expand them.  May be passed more than once.
+    Currently only supported when also using :option:`--aws-batch`.
+    Default is to upload the entire pathogen build directory (except
+    for some ancillary files which are always excluded).
+
+    Note that files excluded from upload may still be downloaded from
+    the remote build, e.g. if they're created by it, and if downloaded
+    will overwrite the local files.  Use :option:`--no-download` to
+    avoid downloading any files, or :option:`--exclude-from-download`
+    to avoid downloading specific files, e.g.::
+
+        --exclude-from-upload 'results/**' \
+        --exclude-from-download 'results/**'
+
+    Your shell's brace expansion can also be used to shorten this, e.g.::
+
+        --exclude-from-{up,down}load='results/**'
+
+    Besides basic glob features like single-part wildcards (``*``),
+    character classes (``[…]``), and brace expansion (``{…, …}``),
+    several advanced globbing features are also supported: multi-part
+    wildcards (``**``), extended globbing (``@(…)``, ``+(…)``, etc.),
+    and negation (``!…``).
+
+    Patterns should be relative to the build directory.
+
+
+
 
 .. option:: --no-logs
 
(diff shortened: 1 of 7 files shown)


tsibley:
Ah, this has a bug around absolute vs. relative path handling and pattern specification. I'll address it and repush.

tsibley:
Bug squashed in [repush](https://github.com/nextstrain/cli/compare/8bc16320effbed0fcdfceab33aec560da0dba9d3..6d465f0dc9f2ca1f7384b948b76661844584a6f1).

doc/commands/build.rst
```diff
@@ -97,13 +97,67 @@ options
...
+    character classes (``[…]``), and brace expansion (``{…, …}``),
+    several advanced globbing features are also supported: multi-part
+    wildcards (``**``), extended globbing (``@(…)``, ``+(…)``, etc.),
+    and negation (``!…``).
+
+    Patterns should be relative to the build directory.
+
+
+
+
```
genehack, line 160: <NIT> the amount of vertical whitespace isn't consistent between options. </NIT>

genehack (approved):

doc/commands/build.rst
tsibley: Yeah, good 👀s ;-)

These doc files are entirely generated (via `devel/generate-command-doc`). The extra space is due to how we skip the default behaviour of showing the option default value in the option help:

https://github.com/nextstrain/cli/blob/6d465f0dc9f2ca1f7384b948b76661844584a6f1/nextstrain/cli/command/build.py#L176-L179

It leaves trailing newlines as a small side-effect.

Could we clean it up with a well-placed `.rstrip()` or an uglier interpolation of `SKIP_AUTO_DEFAULT_IN_HELP`? Yeah. But it's not a big concern.

tsibley:
@lmoncla @jameshadfield @trvrb If you want to test this out before merge/release to make sure it works for you, you can temporarily install the CLI built from this PR with:

    curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac \
      | DESTINATION=/tmp/cli bash -s pr-build/370

Ignore the shell setup instructions printed at the end. To test, run `/tmp/cli/nextstrain build …` instead of `nextstrain build …`.

doc/commands/build.rst
```diff
@@ -97,13 +97,67 @@ options
     wildcards (``**``), extended globbing (``@(…)``, ``+(…)``, etc.),
     and negation (``!…``).
 
+    Patterns should be relative to the build directory.
```
jameshadfield, line 100: _Potentially tangential to this PR_

Reading this code my understanding is that only the build directory is uploaded, and so patterns are relative to that. How does this work when the build directory is (e.g.) `ingest/` but within that the Snakefile refers back to a parent directory?

jameshadfield (approved):
Looks good by inspection. Nice to see the implementation wasn't too far from [how I thought](https://github.com/nextstrain/cli/issues/219#issuecomment-1248808596).

doc/commands/build.rst
tsibley: Ah, that's a great point. The directory here that matters—the one that's uploaded and one that patterns must be relative to—is influenced by the `nextstrain-pathogen.yaml` file. For example, running `nextstrain build ingest/` in a directory with `nextstrain-pathogen.yaml` will require patterns be relative to the directory containing the marker file, not the `ingest/` directory, i.e. to exclude something in `ingest/` would require the pattern to be `ingest/x/y/z` not just `x/y/z`.

I'll update the docs here.

tsibley: Actually, I want to leave the option docs/help as-is for now for this PR. I'm realizing there's larger documentation reworking to do to clarify terminology here as a result of https://github.com/nextstrain/cli/pull/355 (something I'd postponed in that PR). Writing a clear help snippet for these pattern options will be much easier and less inconsistent/confusing once the larger terminology shift and doc updates happen. I should prioritize that.

tsibley: https://github.com/nextstrain/cli/issues/371

jameshadfield: 👍  totally forgot about this subtlety. Is this behaviour mirrored in what directories are bind-mounted into the docker container?

tsibley: Yep.

Before #355 the directory given to `nextstrain build` ended up as `/nextstrain/build` in the container and `/nextstrain/build` was the fixed working directory for the container process.

After #355, the `/nextstrain/build` mount is either the directory given to `nextstrain build` or, if present, the closest ancestor of the given directory which contains a `nextstrain-pathogen.yaml`. The working directory for the container process is the `/nextstrain/build` path corresponding to the directory given to `nextstrain build`.

tsibley:
This feature will be in the being-released-as-we-speak 8.4.0.

Issue #382, opened 2024-07-24
CI on macOS failing due to IQTREE


tsibley:
IQTREE is [triggering](https://github.com/nextstrain/cli/actions/runs/10081483219/job/27873611711#step:6:1889) an illegal instruction:

```
ERROR: Shell exited from fatal signal SIGILL when running: iqtree2 -ntmax 1 -s results/aligned-delim.fasta -m GTR -ninit 2 -n 2 -me 0.05 -nt AUTO -redo > results/aligned-delim.iqtree.log
```

tsibley:
This seems persistent, so I suspect some issue with an IQ-TREE version change in Bioconda, which is used to setup and test an ambient runtime for CI.

Notably, the [last successful CI run](https://github.com/nextstrain/cli/actions/runs/10065919443) on 23 July [got `bioconda/osx-64::iqtree==2.3.5==h63051a7_1`](https://github.com/nextstrain/cli/actions/runs/10065919443/job/27826119640#step:3:945). The [first failing CI run](https://github.com/nextstrain/cli/actions/runs/10081483219), just the next day on 24 July, [got `bioconda/osx-64::iqtree==2.3.5==h49642b6_2`](https://github.com/nextstrain/cli/actions/runs/10081483219/job/27873612537#step:3:945). The build number change from 1 → 2 potentially implicates the [Bioconda recipe change](https://github.com/bioconda/bioconda-recipes/commit/f8218b465339745434a31a35a97329c846c93a44). Both ran on macOS 14 arm64 runners, with the same runner image (`20240714.2`).

tsibley:
```
blab@blab-mac ~ % sw_vers
ProductName:	macOS
ProductVersion:	12.6.8
BuildVersion:	21G725

blab@blab-mac ~ % arch
arm64

blab@blab-mac ~ % CONDA_SUBDIR=osx-64 micromamba create -p /tmp/iqtree -c bioconda -c conda-forge 'bioconda/osx-64::iqtree 2.3.5 h49642b6_2'
Found conda-prefix at '/private/tmp/iqtree'. Overwrite?: [y/N] y
bioconda/osx-64                                             Using cache
bioconda/noarch                                             Using cache
conda-forge/osx-64                                          Using cache
conda-forge/noarch                                          Using cache

Transaction

  Prefix: /private/tmp/iqtree

  Updating specs:

   - bioconda/osx-64::iqtree==2.3.5=h49642b6_2


  Package        Version  Build       Channel           Size
[...]

PR #384, opened 2024-07-29
dev: Upgrade setup-miniconda action from v2 → v3

### LINKED ISSUE #386: CI: deprecation warning for node 16 use in setup-miniconda v2
We're getting a lot of warnings in CI summary due to use of setup-miniconda v2

> The following actions uses Node.js version which is deprecated and will be forced to run on node20: conda-incubator/setup-miniconda@v2. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

from: https://github.com/nextstrain/cli/actions/runs/10183163639

Google Chrome 2024-07-31 18 22 40

tsibley:
The latest version will now use Conda's osx-arm64 channel on macOS when available.  As it seems we're encountering issues with osx-64 packages on arm64 hardware¹, let's see if this switch helps avoid those issues.

¹ <https://github.com/nextstrain/cli/issues/382>

Resolves #386 

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/actions/setup-integration-tests/action.yaml
+++ b/.github/actions/setup-integration-tests/action.yaml
@@ -22,7 +22,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: conda-incubator/setup-miniconda@v2
+    - uses: conda-incubator/setup-miniconda@v3
       with:
         python-version: ${{ inputs.python-version }}
         miniforge-variant: Mambaforge


tsibley:
Ok, so this change did cause Conda's osx-arm64 packages to get used during setup-integration-tests. I expected that would fix the illegal instruction issue in the ambient runtime portion of run-integration-tests. It might! but CI failed earlier than that in the Conda runtime portion. In retrospect, this makes sense, as the Conda runtime is currently osx-64 only and it incorporates the bad iqtree build as of [nextstrain-base 20240724T185322Z](https://github.com/nextstrain/conda-base/actions/runs/10082385556#summary-27876502013). I confirmed that the first iqtree illegal instruction failure seen in the ambient runtime set up for CI was in a run with an older nextstrain-base. So to fix this we'll have to update nextstrain-base too, either by excluding the bad build or (presumably) releasing an osx-arm64 version of it too.

corneliusroemer:
I retriggered CI as failure was due to now fixed #382

corneliusroemer (approved):

tsibley:
Repushed with an updated commit message to reflect current reality and merged.

image

PR #404, opened 2024-10-21
dev: Update GitHub runners used by CI

tsibley:
macOS 12 was deprecated in August and will be removed by December.¹  I moved the build and test jobs to the next oldest version, macOS 13.

Ubuntu 24.04 was released in late September,² so test on it.

macOS 15 was released for beta in late September,² so test on it too.

Test the x86_64 binary on macOS 14 and 15 too since it should work with Rosetta.  I'm not sure why it wasn't tested on macOS 14 previously; maybe just an oversight?

¹ <https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/>
² <https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/>

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/actions/setup-integration-tests/action.yaml
+++ b/.github/actions/setup-integration-tests/action.yaml
@@ -91,7 +91,11 @@ runs:
         url="$(
           curl -fsSL --proto '=https' -H "Authorization: Bearer $GITHUB_TOKEN" \
             "https://api.github.com/repos/sylabs/singularity/releases?per_page=100" \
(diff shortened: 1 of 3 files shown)


tsibley:
[Failed](https://github.com/nextstrain/cli/actions/runs/11447349955/job/31848998402?pr=404#step:3:1379) test-standalone job on ubuntu-24.04 because SingularityCE doesn't publish packages specifically for that Ubuntu release ("noble"). The download URL is emitted as `null` from jq, which the second curl invocation treats as a bare hostname, e.g. `http://null`, hence the "Protocol "http" not supported or disabled in libcurl" error message.

As a fix, I can adjust `setup-integration-tests` to use next closest Ubuntu release name.

victorlin (approved):

genehack (approved):

PR #418, opened 2025-02-13
Unbuffer runner test results

### LINKED ISSUE #359: Show progress during `check-setup`
`nextstrain check-setup` takes almost a minute despite having a fast M1 Pro. There's no progress indicator which might be confusing to users who are running it for the first time.

It might be good to give an indication how long this check takes. E.g. "Testing your setup... (this takes about a minute)"

Even better would be some sort of progress indication after each runtime is checked, maybe outputting it or indicating that it has been checked successfully.

image

victorlin:

setup and check-setup commands can take several seconds for certain runtimes. Improve user experience by showing progress as it happens.

## Related issue(s)

Closes #359

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -38,7 +38,7 @@ from functools import partial
 from .. import config
 from ..argparse import SKIP_AUTO_DEFAULT_IN_HELP, runner_module_argument
 from ..types import Options
-from ..util import colored, check_for_new_version, runner_name, runner_tests_ok, print_runner_tests
+from ..util import colored, check_for_new_version, runner_name, runner_tests_ok, passthru_and_print_runner_tests
 from ..runner import all_runners, all_runners_by_name, default_runner # noqa: F401 (it's wrong; we use it in run())
 
 
@@ -84,37 +84,32 @@ def run(opts: Options) -> int:
     # Run and collect our runners' self-tests
     print("Testing your setup…")
 
-    runner_tests = [
+    runner_tests = (
         (runner, runner.test_setup())
             for runner in opts.runners
-    ]
+    )
 
-    runner_status = {
-        runner: runner_tests_ok(tests)
-            for runner, tests in runner_tests
-    }
+    supported_runners = []
 
     # Print test results.  The first print() separates results from the
     # previous header or stderr output, making it easier to read.
     print()
 
     for runner, tests in runner_tests:
-        if runner_status[runner]:
+        print(colored("blue", "#"), "Checking %s…" % (runner_name(runner)))
+
+        tests = passthru_and_print_runner_tests(tests)
+
+        if runner_tests_ok(tests):
             supported = success("supported")
+            supported_runners.append(runner)
         else:
             supported = failure("not supported")
 
         print(colored("blue", "#"), "%s is %s" % (runner_name(runner), supported))
-        print_runner_tests(tests)
         print()
 
     # Print overall status.
-    supported_runners = [
-        runner
-            for runner, status_ok in runner_status.items()
-             if status_ok
-    ]
-
     if supported_runners:
         print("Supported Nextstrain runtimes:", ", ".join(success(runner_name(r)) for r in supported_runners))
 
--- a/nextstrain/cli/command/check_setup.py
+++ b/nextstrain/cli/command/check_setup.py
@@ -38,7 +38,7 @@ from functools import partial
 from .. import config
 from ..argparse import SKIP_AUTO_DEFAULT_IN_HELP, runner_module_argument
 from ..types import Options
-from ..util import colored, check_for_new_version, runner_name, runner_tests_ok, passthru_and_print_runner_tests
+from ..util import colored, check_for_new_version, runner_name, print_and_check_runner_tests
 from ..runner import all_runners, all_runners_by_name, default_runner # noqa: F401 (it's wrong; we use it in run())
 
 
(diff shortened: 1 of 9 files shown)


tsibley:
Rebased onto latest master for tests.

nextstrain/cli/util.py
```diff
@@ -598,10 +598,12 @@ def runner_tests_ok(tests: RunnerTestResults) -> bool:
     return False not in [result for test, result in tests]
 
 
-def print_runner_tests(tests: RunnerTestResults):
+def stream_and_print_results(tests: RunnerTestResults) -> RunnerTestResults:
```
tsibley, line 601: I think "passthru" is the more relevant word than "stream": the input is (passed thru to) the output.

Also, "results" is left unqualified in the new function name. What results? Whose results?

```suggestion
def passthru_and_print_runner_tests(tests: RunnerTestResults) -> RunnerTestResults:
```

nextstrain/cli/util.py
```diff
@@ -628,6 +630,8 @@ def print_runner_tests(tests: RunnerTestResults):
 
         print(status.get(result, str(result)) + ":", formatted_description)
 
+        yield (description, result)
```
tsibley, line 633: Move this above the `print()` (to first line of loop body probably) to make fully true the docstring claim:

> Messages will only be printed if the generator is consumed downstream.

(as otherwise the first message will always be printed) and to also match the order of operations implied by the function name.

nextstrain/cli/runner/conda.py
```diff
@@ -422,30 +422,27 @@ def runnable(*argv) -> bool:
 
     support = test_support()
 
+    yield from support
+
     if not runner_tests_ok(support):
-        return support
+        pass
```
tsibley, line 428: In the context of the switch to generators, this re-use of `support` surprised me and looked like a bug. It's not because `test_support()` is still returning a list instead of switching to a generator. This is a bug waiting to happen though, e.g. if someone goes, "Oh, `test_support()`'s not a generator yet, I'll fix that." I'd suggest the following small change to more can guard against that and more clearly indicate expectations.

```suggestion
    support = list(test_support())

    yield from support

    if not runner_tests_ok(support):
        return
```

(includes using `return` instead of `pass` to return early, matching existing code)

nextstrain/cli/command/setup.py
```diff
@@ -70,9 +70,9 @@ def run(opts: Options) -> int:
     print(heading(f"Checking setup…"))
 
     if not opts.dry_run:
-        tests = opts.runner.test_setup()
+        tests: RunnerTestResults = opts.runner.test_setup()
 
-        print_runner_tests(tests)
+        tests = stream_and_print_results(tests)
 
         if not runner_tests_ok(tests):
```
tsibley, line 77: _non-blocking_

This API is a little weird. I'd wonder about something more like 

```python
ok = print_and_check_runner_tests(tests)

if ok:
```

by having `print_and_check_runner_tests()` buffer internally and `return runner_tests_ok(tests)`.

tsibley (changes requested):
+1 for functionality, a few comments on details of implementation.

nextstrain/cli/command/setup.py
victorlin: Updated with 0597de130942e1a66b238b992def830d660e3083

nextstrain/cli/command/check_setup.py
```diff
@@ -84,37 +84,33 @@ def run(opts: Options) -> int:
...
     print()
 
     for runner, tests in runner_tests:
-        if runner_status[runner]:
+        print(colored("blue", "#"), "Checking %s…" % (runner_name(runner)))
+
+        tests = stream_and_print_results(tests)
+
+        runner_status = runner_tests_ok(tests)
+        if runner_status:
```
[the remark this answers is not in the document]
victorlin: Done:

https://github.com/nextstrain/cli/blob/f6ddf6e139fd56aa63568955901c11786717e0fd/nextstrain/cli/command/check_setup.py#L103

and also changed in https://github.com/nextstrain/cli/pull/418#discussion_r2010958707

nextstrain/cli/runner/conda.py
victorlin: Good catch, done:

https://github.com/nextstrain/cli/blob/f6ddf6e139fd56aa63568955901c11786717e0fd/nextstrain/cli/runner/conda.py#L423

I've also updated `test_support()` and everything else that returns `RunnerTestResults` to return a generator.

https://github.com/nextstrain/cli/blob/f6ddf6e139fd56aa63568955901c11786717e0fd/nextstrain/cli/runner/conda.py#L467-L473

nextstrain/cli/util.py
victorlin: Updated to `passthru_and_print_runner_tests`:

https://github.com/nextstrain/cli/blob/f6ddf6e139fd56aa63568955901c11786717e0fd/nextstrain/cli/util.py#L601

then changed to `print_and_check_runner_tests` for https://github.com/nextstrain/cli/pull/418#discussion_r2010958707

nextstrain/cli/util.py
victorlin: Done:

https://github.com/nextstrain/cli/blob/f6ddf6e139fd56aa63568955901c11786717e0fd/nextstrain/cli/util.py#L631-L633

though order matters less now with https://github.com/nextstrain/cli/pull/418#discussion_r2010958707:

https://github.com/nextstrain/cli/blob/0597de130942e1a66b238b992def830d660e3083/nextstrain/cli/util.py#L631-L633

tsibley (approved):

Issue #328, opened 2023-11-08
Consider UI/UX for AWS Batch runtime without Docker runtime


joverlee521:
We have not considered users who are using the AWS Batch runtime but not the Docker runtime, but it's [definitely possible](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1699314296825929?thread_ts=1699299368.404799&cid=C7SDVPBLZ)!

I think the main point to consider here is how to manage the image used for AWS Batch. Maybe we should consider adding a `nextstrain update aws-batch` command that can update the docker image version? Maybe add image to the AWS Batch config section? I think either would add more complexity to https://github.com/nextstrain/cli/issues/167...

tsibley:
Good to consider this. I think the main decision point is if the AWS Batch runtime image should stay tightly coupled to the Docker runtime image.

If it should, then `nextstrain update aws-batch` (and `nextstrain setup aws-batch`) could touch the `docker.image` config var. But this might be somewhat unintuitive and surprising, as other runtimes are independent of each other.

It would be more consistent to decouple the AWS Batch and Docker runtime images so they can be setup/updated separately. But this would also be a behavioural change which itself may surprise folks who expect them (consciously or not) to be synced.

tsibley:
Steph from WA DOH recently [tripped over this](https://bedfordlab.slack.com/archives/C01G4B4UY31/p1730137919847849). She uses our Conda runtime locally but also our AWS Batch runtime and had previously used our Docker runtime locally so had an old image version in config.

tsibley:
James [reports](https://bedfordlab.slack.com/archives/C01G4B4UY31/p1730144829950559?thread_ts=1730137919.847849&cid=C01G4B4UY31) being tripped up by this in the past too.

PR #419, opened 2025-03-05
Grab bag of bug fixes and improvements

tsibley:
All related of varying degrees to my `nextstrain run` and workflows-as-programs work in #407. Split out to be reviewed/merged (and probably released) separately. One big grab bag, as [no one](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1741119826146619) seemed to mind.

## Checklist

- [ ] Checks pass


CHANGES.md
jameshadfield, line 1: > Extending overlay support to AWS Batch has been very low priority and something I thought was unlikely to ever happen.

I've wanted this! Awesome that it's happening

> One caveat compared to overlays in runtimes with the concept of volume mounts (Docker, Singularity) is that any files in the image that do not exist in the overlaid files will remain present since nothing removes them.  This is potentially problematic and will be annoying if run into but most of the time should be a non-issue.  It is also solvable if we care to exert the effort and extra code to do so.  I don't right now.

This'll be a nightmare to debug if it ever occurs. If I understand the code (I don't spend much time in this codebase) the behaviour of `--augur` is going to differ if it's run locally in docker (which swaps the entire `augur` out) vs aws-batch (which overlays any zipped up `../augur` contents), right? Since `--augur` is already special cased I think it'd be sensible to remove the (containers original) `augur` directory if we see `../augur` in the zip file.

nextstrain/cli/runner/aws_batch/s3.py
```diff
@@ -80,8 +83,23 @@ def upload_workdir(workdir: Path, bucket: S3Bucket, run_id: str, patterns: List[
...
+                #   -trs, 10 Feb 2025
+                try:
+                    prefix = PurePath(volume.name).relative_to("build")
+                except ValueError:
+                    prefix = PurePath("..", volume.name)
+
+                for path in walk(volume.src, always_excluded):
+                    dst = prefix / path.relative_to(volume.src)
+                    print(f"zipping: {path}" + (f" (as {dst})" if DEBUGGING else ""))
+                    zipfile.write(str(path), dst)
```
jameshadfield, line 102: Not related to this PR, but for the `nextstrain run` stuff it'd be a good idea to ensure the analysis directory doesn't clash with any existing directories in the image. Or perhaps you're going to unzip it to some hardcoded directory name to avoid this? I think someone will eventually name their analysis directory "augur" or something. (I'm assuming this code is how the CLI will include analysis directories in the zip file)

jameshadfield (commented):
Skimmed over most of the code, but it was the last commit that I was most interested in!

CHANGES.md
tsibley: > I've wanted this! Awesome that it's happening

\o/

> This'll be a nightmare to debug if it ever occurs. If I understand […]

Your understanding is correct. I think "nightmare" is overstating it—a stack trace, for example, will implicate a file that shouldn't be there—but yes, it'd certainly be aggravating.

Note that `--augur` is just one example of these development overlays, although probably the most useful one. There's also `--auspice`, `--fauna`, and (ha) `--sacra`. (I've personally used `--augur` and `--auspice` about evenly.)

Additional handling of these overlays in the AWS Batch entrypoint is possible, like I noted, but I don't really want to spend more time on those bits now. The overlay options have other caveats that present tripping hazards too, e.g. for Augur, it's just source code that's overlaid, not dependencies; and for Auspice you need to make sure to build it locally first. So a bit of "here be dragons" isn't unprecedented, and these are (AFAIK) rarely-used options.

jameshadfield: Well, if anything comes from this discussion we should at least remove `--sacra`

jameshadfield: As a little background, it's very very rare I run a workflow using the release version of augur + the `master/main` repo branch, as that's what we have the automated stuff for in my eyes. Almost every time I'm running workflows it's because i've modified the workflow, or modified augur, or (surprisingly often) both. If we can get `nextstrain run ...` working as I hope so I can farm out a job to AWS with my modified augur code & modified repo and then download the assets as a separate siloed working directory I'd save a whole lot of build time and my laptop would be happier. Is #407 at the point where this can be done / tested out?

tsibley: This is a good reminder that although I often focus on the expected/anticipated external users of `nextstrain run`, we have internal uses for it that have different focuses. Internally we care mainly about the workdir isolation and mechanics in combination with offloading the work to AWS Batch, not the pathogen source/version management.

You can totally test out #407 for your purposes!

There is not currently a supported way to use `nextstrain setup` to make an "editable install" (to use Python/Pip's lingo) of a self-managed local pathogen repo. It's on the list, but I wasn't focusing on it at first. Sounds like it should be soon/next up. In the meantime, a workaround is to do something like this:

```console
$ ln -sv ~/nextstrain/measles ~/.nextstrain/pathogens/measles/local="$(echo -n local | base32)"
$ nextstrain run measles@local …
```

jameshadfield: > This is a good reminder that although I often focus on the expected/anticipated external users

Yeah, this'll always be a tension in the group - imaging a "happy path" for "most people" vs the way some of us run things. For this work I don't mind symlinking things.

P.S. `base32` not available on MacOS out of the box.

nextstrain/cli/command/build.py
```diff
@@ -306,10 +306,31 @@ def assert_overlay_volumes_support(opts):
...
+                {opts.image}
+
+            is too old to support overlays (e.g. of {overlay_volumes[0].name}) with AWS Batch.
+
+            If overlays are necessary, please update the runtime image to at
+            least version
+
+                {split_image_name(opts.image)[0]}:{docker.IMAGE_FEATURE.aws_batch_overlays.value}
+
+            using `nextstrain update docker`.  Alternatively, use the Docker or
```
joverlee521, line 332: Seeing `nextstrain update docker` for the `aws_batch` runner reminded me of https://github.com/nextstrain/cli/issues/328 😅 

Maybe include `--image <version>` in the error message as an alternative.

nextstrain/cli/runner/docker.py
joverlee521, line 1: _very minor nit_ 

It's a little weird to me that the overlay options (e.g. `--augur`) are still housed within `runner/docker.py` now that it can be used by multiple runners. (The [TODO comment](https://github.com/nextstrain/cli/blob/9d6a083cac6fc9a587e6d24ee1f41bd3fdf0c735/nextstrain/cli/runner/docker.py#L119-L125) is now outdated). 

I suggest moving it to `runner/__init__.py`, under [the --image option](https://github.com/nextstrain/cli/blob/9d6a083cac6fc9a587e6d24ee1f41bd3fdf0c735/nextstrain/cli/runner/__init__.py#L189-L195).

joverlee521 (approved):
Overall changes LGTM. I only left minor comments.

nextstrain/cli/command/build.py
tsibley: Yeah, same… it'd be good to address that.

I intentionally left out `--image` to a) not further complicate the message and b) gently push people towards updating since periodically updating is more broadly preferable than slapping in `--image` at a fixed version and never ever updating it. I figure people already familiar with `--image` will realize they can use it too.

If you think otherwise, I'll add it in, though.

tsibley: (Probably should have written that in a comment in the code to start…)

CHANGES.md
```diff
@@ -29,6 +56,41 @@ development source code and as such may not be routinely kept up to date.
   in a browser (if any).
   ([#417](https://github.com/nextstrain/cli/pull/417))
 
+* `nextstrain login` now correctly handles OAuth 2.0 authorization endpoint
+  URLs (which are obtained automatically from OpenID Connect 1.0 metadata
+  discovery) with existing query parameters in them.  This bug likely effected
```
genehack, line 61: _nit_
```suggestion
  discovery) with existing query parameters in them.  This bug likely affected
```

genehack (approved):

nextstrain/cli/runner/docker.py
tsibley: Yeah, for sure. That's a good suggestion. Done in 1a27583.

CHANGES.md
tsibley: Ugh, thanks, fixed in 890e3d0.

CHANGES.md
tsibley: > @jameshadfield: Since `--augur` is already special cased I think it'd be sensible to remove the (containers original) `augur` directory if we see `../augur` in the zip file.

> @tsibley: Additional handling of these overlays in the AWS Batch entrypoint is possible, […] but I don't really want to spend more time on those bits now.

Following up on this, new week new me and I'm maybe coming around to JFDI to head off future headache.

Since it'll definitely be done on the container entrypoint side and likely without any need for changes to Nextstrain CLI itself, this means it'll be fixed by an image update. So I don't see reason to wait on it for this PR, even if we are decided that it's a definitely now thing and not a maybe later thing. I'll likely tinker a bit on the image side and see how annoying it is to implement after all.

tsibley: > Well, if anything comes from this discussion we should at least remove `--sacra`

86fb154

nextstrain/cli/command/build.py
joverlee521: Yeah, it just feels a little weird to suggest a command that would not work for a subset of users that don't use the docker runtime locally. If the user does not use the docker runtime then I _think_ the only way they would have requested an outdated image is via `--image` or `NEXTSTRAIN_DOCKER_IMAGE`.

For not wanting to complicate the message, maybe just remove the "using `nextstrain update docker`." bit.

tsibley: Oh I see. I missed/forgot that it wasn't just weird but it'd actually just not work. Right.

tsibley: I want to either fully separate or fully unify the image versions used between runtimes (i.e. make AWS Batch uncoupled from Docker, or more tightly couple versions used by Docker/Singularity/AWS Batch), but since that's not happening immediately, I decided to [expand](https://github.com/nextstrain/cli/compare/86fb15446ad34d36e8655383ce926f46b5e41946..2837318f7ac44c1ad8fd65325076acef9c3eb11a) the error message a bit.

Thanks for helping me see the issue here!

CHANGES.md
tsibley: > I'll likely tinker a bit on the image side and see how annoying it is to implement after all.

Looks like the actual functional change might be this line:

```diff
diff --git a/entrypoint-aws-batch b/entrypoint-aws-batch
index 7e693c2..41bae21 100755
--- a/entrypoint-aws-batch
+++ b/entrypoint-aws-batch
@@ -8,6 +8,7 @@ set -x
 case "$NEXTSTRAIN_AWS_BATCH_WORKDIR_URL" in
     s3://*.zip)
         aws s3 cp --no-progress "$NEXTSTRAIN_AWS_BATCH_WORKDIR_URL" "$PWD.zip"
+        zipinfo -1 "$PWD.zip" | grep --extended-regexp --regexp='^[.][.]/'{augur,auspice,fauna}'/?$' || true | xargs --no-run-if-empty rm -rfv
         unzip -: -o "$PWD.zip"
         ;;
     s3://*)
```

tsibley: https://github.com/nextstrain/docker-base/pull/245 addresses the overlay caveat.

tsibley: https://github.com/nextstrain/cli/pull/423 makes it required.

jameshadfield: Thanks! A good reminder to me of where the moving pieces are in this puzzle

PR #431, opened 2025-04-25
pathogens: Resolve versions via a new nextstrain.org API

tsibley:
See commit messages.

## Checklist

- [x] Checks pass

### DIFF
--- a/nextstrain/cli/pathogens.py
+++ b/nextstrain/cli/pathogens.py
@@ -26,7 +26,7 @@ from .errors import UserError
 from .net import is_loopback
 from .paths import PATHOGENS
 from .types import SetupStatus, SetupTestResults, SetupTestResult, UpdateStatus
-from .url import URL
+from .url import URL, NEXTSTRAIN_DOT_ORG
 from .util import parse_version_lax, print_and_check_setup_tests, request_list
 
 
@@ -189,7 +189,7 @@ class PathogenVersion:
 
         if not version:
             if new_setup:
-                version = github_repo_latest_ref(f"nextstrain/{name}")
+                version = nextstrain_repo_latest_version(name)
             else:
                 version = default_version_of(name, implicit = True)
 
@@ -281,7 +281,7 @@ class PathogenVersion:
 
         if new_setup:
             if not url:
-                url = github_repo_ref_zipball_url(f"nextstrain/{name}", version)
+                url = nextstrain_repo_zip_url(name, version)
 
             if not url:
                 raise UserError(f"""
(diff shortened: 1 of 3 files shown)


genehack (approved):

nextstrain/cli/pathogens.py
```diff
@@ -845,62 +856,54 @@ def read_pathogen_registration(path: Path) -> Dict:
...
     """
-    Queries a GitHub *repo_name* to return the name of the highest version tag,
-    if any, otherwise the name of the default branch.
+    Queries a Nextstrain pathogen repo *name* to return the latest version,
+    i.e. the name of the highest version tag, if any, otherwise the name of the
+    default branch.
     """
+    org = URL(NEXTSTRAIN_DOT_ORG)
+
+    versions_url = URL(f"https://nextstrain.org/pathogen-repos/{urlquote(name)}/versions")._replace(scheme = org.scheme, netloc = org.netloc)
```
joverlee521, line 876: I'm confused by the split of `org` and `versions_url` here...I thought  25e6560a3e30780a8f6012149a25a678c83b54f9 would allow adding the base to the URL:

```suggestion
versions_url = URL(f"pathogen-repos/{urlquote(name)}/versions", NEXTSTRAIN_DOT_ORG)
```

joverlee521 (approved):
LGTM. Tested with 
```
NEXTSTRAIN_DOT_ORG="https://nextstrain-s-trs-pathog-aosx4t.herokuapp.com/" nextstrain setup measles@main
```

Left a non-blocking question for my own understanding.

nextstrain/cli/pathogens.py
tsibley: Ah, good question! There's indeed a subtly here.

We want to take only the scheme and netloc (host:port) from `NEXTSTRAIN_DOT_ORG`, to be consistent with how the override is used by `nextstrain remote` and `nextstrain login` commands, e.g.

https://github.com/nextstrain/cli/blob/8007c4790c55ebb17f452956c4e530d479bd2dff/nextstrain/cli/remote/__init__.py#L123-L125

Using it as a base URL, as in `URL(…, NEXTSTRAIN_DOT_ORG)`, would also allow potential path segments to be taken/used as well, e.g.

```python-console
>>> str(URL(f"a/b/c", "https://example.com/x/y/z/"))
'https://example.com/x/y/z/a/b/c'
```

joverlee521: Ah, I see. Thanks for the explanation!

PR #432, opened 2025-04-28
User agent

tsibley:
See commit messages.

## Checklist

- [ ] Checks pass


genehack (approved):

tsibley:
I started work server-side to parse out some of this info for analytics, and I'm now planning to adjust the user-agent format a little bit here to make it easier to parse out. I also plan to drop the Python implementation info, which I think is unnecessary in retrospect.

PR #433, opened 2025-04-28
util: Update PyPI URL to respect 301 Permanent Redirect

tsibley:
When using mitmproxy to observe other HTTP traffic during other development, I noticed that the pypi.python.org URL we were using was responding with a 301 to pypi.org.  It'd be sweet as if requests locally remembered 301s (like a browser does), but it doesn't, and I'm not about to add that feature right now.

## Checklist

- [ ] Checks pass


genehack (approved):

tsibley:
Replaced by https://github.com/nextstrain/cli/pull/434.

PR #434, opened 2025-04-30
util: Fetch our own latest version via nextstrain.org instead of pypi.org

tsibley:
This gives us insights into usage and also more flexibility to shift how we're releasing and distributing it.

Related-to: <https://github.com/nextstrain/nextstrain.org/pull/1160>

Replaces/incorporates #433.

## Checklist

- [ ] Checks pass


tsibley:
CI tests will fail until https://github.com/nextstrain/nextstrain.org/pull/1160 is merged, deployed, and promoted to nextstrain.org.

genehack (approved):

PR #438, opened 2025-05-29
docs: Include a changelog in the documentation

tsibley:
This has been a longstanding omission that I finally got around to correcting because I wanted to more publicly link to the changelog and not point people into the den of source-ry that is GitHub.

I did this thing not because it was easy, but because I *thought* it would be easy: just drop in CHANGES.md as-is using an "include" directive. Well, after much gnashing of keys, it turns out that usable heading anchors are impossible to achieve with this method unless we make CHANGES.md include noisy MyST syntax. So I swapped to doc generation/transformation instead.

## Checklist

- [x] Checks pass

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,6 +4,9 @@ build:
   os: ubuntu-20.04
   tools:
     python: "3.10"
+  jobs:
+    pre_build:
+      - ./devel/rtd-pre-build
 sphinx:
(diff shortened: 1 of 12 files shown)


tsibley:
Merging so I can link to https://docs.nextstrain.org/projects/cli/en/latest/changes/#v-10-0-0 elsewhere. I'd welcome post-review feedback, though!

PR #441, opened 2025-06-11
Update AWS Batch docs to reflect latest AWS console interface

victorlin:

I just ran through this setup guide on the Bedford lab's Hutch-managed AWS account. Some parts were confusing with the current AWS console interfaces. My guess is that some things got moved around since this guide was written. This PR updates the guide to reflect the UI that I just used.

## Checklist

- [x] Checks pass

### DIFF
--- a/doc/aws-batch.md
+++ b/doc/aws-batch.md
@@ -336,8 +336,8 @@ Specify the container image `nextstrain/base:latest` and an empty command.
 (In the wizard, delete the pre-filled command, leaving the JSON result as an
 empty array (`[]`).)
 
-Choose the job role _NextstrainJobsRole_, which you just created in the IAM
-roles section above.
+Do not choose any execution role. For the job role, choose _NextstrainJobsRole_
+which you just created in the IAM roles section above.
 
 Select the number of desired vCPUs and amount of memory you'd like each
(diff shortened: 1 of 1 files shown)


doc/aws-batch.md
```diff
@@ -404,17 +404,20 @@ instance.
...
+
+> Note: if your compute environment uses the launch template with version set to
+`$Latest`, you must create a new compute environment to apply any new changes to
+the launch template.  Compute environments set to use the `$Latest` version of a
+launch template are frozen to the latest template version that exists at the
+time the environment was created, per [AWS Batch documentation][compute
+environment launch template].  For this reason, it's recommended to use an
+explicit version number instead of `$Latest` so that you can easily see what
+version a compute environment is using (instead of having to correlate compute
+environment and launch template version creation times).
```
tsibley, line 420: > I was able to set this up successfully with an existing compute
environment. My understanding, based on the message which is now in a
note, is that creating a new compute environment is only necessary when
the existing environment was created with the launch template version
set to `$Latest`, which won't happen for users following this guide from
scratch.

Oh, hmm. I think you used to be unable to modify a compute environment's launch template version, regardless of if it used a specific version number or `$Latest`. Seems like that's not the case any more?

Issue #269, opened 2023-03-27
Add `self` to `nextstrain update` allowed arguments


corneliusroemer:
### Context
When updating `nextstrain` runtimes, I was told that a new cli version is available and that I should update using:

```
mamba install -p /opt/homebrew/Caskroom/miniforge/base/envs/nextstrain nextstrain-cli==6.2.1
```

That doesn't actually work:

```console
nextclade on  feat/masked-sites-for-placement [$?] via  v18.9.0 via 🦀 v1.67.1 
❯     mamba install -p /opt/homebrew/Caskroom/miniforge/base/envs/nextstrain nextstrain-cli==6.2.1

Looking for: ['nextstrain-cli==6.2.1']

bioconda/osx-arm64                                          Using cache
bioconda/noarch                                             Using cache
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
pkgs/main/osx-arm64                                         Using cache
pkgs/main/noarch                                            Using cache
pkgs/r/osx-arm64                                            Using cache
pkgs/r/noarch                                               Using cache

Pinned packages:
  - python 3.10.*
  - biopython 1.80.*


Could not solve for environment specs
The following package could not be installed
└─ nextstrain-cli 6.2.1  does not exist (perhaps a typo or a missing channel).
```

(but that's not what this issue is about)

### Description
I was wondering whether it would make sense to add `self` update to the options of `nextstrain update`.

Rather than prompting the user to run a specific command (that may actually not work) why not run some code that we know works to update the version?

tsibley:
`mamba install` was recommended as the update mechanism because the copy of `nextstrain` that you're using was installed via the Conda package originally and you have `mamba` available.

The 6.2.1 version was recommended because Nextstrain CLI only checks PyPI for the latest version.

The reason (as I suspect you know, but saying for posterity) that the `mamba install` command didn't work is because the Bioconda nextstrain-cli package [has not yet gotten the 6.2.1 update](https://github.com/bioconda/bioconda-recipes/pull/40056).

The window of time between the PyPI package updating and the Conda package updating is usually relatively short, but I suppose to avoid it we could query the Bioconda package version if we know we're going to recommend updating with it.

[...]

PR #470, opened 2025-09-09
Relocate update message [#457]

### LINKED ISSUE #457: `nextstrain update` behavior is user-hostile
### Current Behavior
When I run `nextstrain update conda` (and, I-assume-but-have-not-tested `nextstrain update docker`) with an outdated version of Nextstrain CLI, at the top of the output, we say this:

```
Upgrade your standalone installation by running:

    curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac | bash -s 10.2.1.post1

or by downloading a new archive from:

    https://nextstrain.org/cli/download/10.2.1.post1/standalone-arm64-apple-darwin.tar.gz
```

…and then if there's actually a `conda-base` update (or, I-assume-but-have-not-tested, a `docker-base` update), we spew several hundred lines of conda output into the terminal, meaning if anybody actually wants to do that update, they either have to do some annoying scrolling, or do what I usually do and re-run the update command again simply to be able to copypasta the `nextstrain-cli` download URL. 

This is especially grating because the `nextstrain conda update` finishes with:

```
…but consider upgrading Nextstrain CLI too, as noted above.
```

### Expected behavior
It would be a much nicer experience if we just repeated that download URL at the bottom of the output, along with the reminder. (Optionally, only repeat it if `conda` / `docker` actually gets updated, because when they don't, the URL remains on screen.)

genehack:

Moves the message about how to update the CLI installation to after all the information about which packages were updated.

## Related issue(s)

Closes #457 
#471 

## Checklist

- [x] Checks pass
- [x] Update changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,12 @@ development source code and as such may not be routinely kept up to date.
   Batch job.
   ([#460](https://github.com/nextstrain/cli/pull/460))
 
+* When Nextstrain CLI detects that a new version of itself is available, it
+  will now output the instructions on how to upgrade to that new version after
+  (instead of before) the normal output from the running command.  This makes
+  it easier to find and follow the instructions when output is long.
+  ([#457](https://github.com/nextstrain/cli/issues/457))
+
 ## Bug fixes
 
 * `nextstrain setup <pathogen>@<version>` and `nextstrain update <pathogen>@<version>`
--- a/nextstrain/cli/command/update.py
+++ b/nextstrain/cli/command/update.py
@@ -76,7 +76,6 @@ def run(opts):
     heading = partial(colored, "bold")
     success = partial(colored, "green")
     failure = partial(colored, "red")
-    notice  = partial(colored, "yellow")
 
     updates: List[Tuple[Callable[[], UpdateStatus], str]] = []
 
@@ -116,7 +115,7 @@ def run(opts):
     # Check our own version for updates
     print(heading(f"Checking for newer versions of Nextstrain CLI…"))
     print()
-    newer_version = check_for_new_version()
+    newer_version, update_instructions = check_for_new_version()
 
     # Perform updates
     if not updates:
@@ -152,14 +151,12 @@ def run(opts):
     # Print overall status
     if all(oks):
         print(success("All updates successful!"))
-        if newer_version:
-            print()
-            print(notice("…but consider upgrading Nextstrain CLI too, as noted above."))
     else:
         print(failure("Some updates failed!  See above for details."))
-        if newer_version:
-            print()
-            print(notice("Maybe upgrading Nextstrain CLI, as noted above, will help?"))
+
+    if newer_version:
+        print()
+        print(update_instructions)
 
     # Return a 1 or 0 exit code
     return int(not all(oks))
--- a/nextstrain/cli/util.py
+++ b/nextstrain/cli/util.py
@@ -73,6 +73,7 @@ def remove_suffix(suffix, string):
 
 def check_for_new_version():
     newer_version = new_version_available()
+    update_instructions = None
 
     installed_into_user_site = \
             site.ENABLE_USER_SITE \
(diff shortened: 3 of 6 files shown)


CHANGES.md
```diff
@@ -28,6 +28,11 @@ development source code and as such may not be routinely kept up to date.
   Batch job.
   ([#460](https://github.com/nextstrain/cli/pull/460))
 
+* When Nextstrain CLI detects that it needs to be updated, it will now
+  output the instructions on how to do any update after the information
+  for which packages were updated.
```
tsibley, line 33: Suggestion to clarify since "update" is used to refer to two different things, and "packages" doesn't apply to runtimes other than Conda or commands like `nextstrain check-setup`.

```suggestion
* When Nextstrain CLI detects that a new version of itself is available, it
  will now output the instructions on how to upgrade to that new version after
  (instead of before) the normal output from the running command.  This makes
  it easier to find and follow the instructions when output is long.
```

nextstrain/cli/util.py
```diff
@@ -129,49 +130,56 @@ def check_for_new_version():
         pkgreq = shquote(f"nextstrain-cli=={newer_version}")
 
         print("A new version of nextstrain-cli, %s, is available!  You're running %s." % (newer_version, __version__))
+        print("See below for upgrade instructions.")
         print()
         print("See what's new in the changelog:")
         print()
         print(f"    https://github.com/nextstrain/cli/blob/{newer_version}/CHANGES.md#readme")
         print()
```
tsibley, line 138: Oh, hmm. I wasn't expecting the split of "new version available" from "here's how to upgrade". I'd think to keep the two together, i.e. all at the bottom. In particular, the changelog link is gonna get lost which really should inform if someone wants to upgrade just yet or not. Was there a reason to keep the notice at the top?

nextstrain/cli/command/update.py
```diff
@@ -116,7 +115,7 @@ def run(opts):
     # Check our own version for updates
     print(heading(f"Checking for newer versions of Nextstrain CLI…"))
     print()
-    newer_version = check_for_new_version()
+    newer_version, update_instructions = check_for_new_version()
 
     # Perform updates
     if not updates:
```
tsibley, line 121: If there are no runtime/pathogen updates to perform, then the update instructions for the CLI itself is never printed.

nextstrain/cli/util.py
```diff
@@ -128,50 +129,58 @@ def check_for_new_version():
...
-        print("See what's new in the changelog:")
-        print()
-        print(f"    https://github.com/nextstrain/cli/blob/{newer_version}/CHANGES.md#readme")
-        print()
+        update_instructions = dedent(f"""
+        A new version of nextstrain-cli, {newer_version}, is available!  You're running {__version__}.
+
+        See what's new in the changelog:
+
+            https://github.com/nextstrain/cli/blob/{newer_version}/CHANGES.md#readme")
```
tsibley, line 137: There's a few trailing `")` left that need trimming off.

tsibley:
(I'm noting these things in comments here to be clearer about what I meant by "a few more things that don't fly", not to make you fix them.)

genehack:
> (I'm noting these things in comments here to be clearer about what I meant by "a few more things that don't fly", not to make you fix them.)

Thanks for saying that, reading these in chrono order I was briefly mystified :grin: 

(Thank you for picking it up; I wasn't terribly pleased with where it ended up but didn't see a better way forward without a whole lot more ditch-digging than this feels like it deserves…)

tsibley:
> without a whole lot more ditch-digging than this feels like it deserves…

*stands up from shoveling in the ditch*

Yeah… it's a surprisingly tricky bit of UI to get right. While I'm in the ditch, I figured I might as well also (finally) make https://github.com/nextstrain/cli/issues/269 go…

tsibley:
I've moved https://github.com/nextstrain/cli/pull/470/commits/0807ac6c9f546605ff379d0692aefa9a969bf947 to https://github.com/nextstrain/cli/pull/473/commits/fa42771a451f6b8b256d0810539fc4be34d92787 as part of #473.

Issue #420, opened 2025-03-12
`nextstrain run`: unmanaged pathogen support


tsibley:
Support `nextstrain run` for unmanaged pathogen sources, i.e. working copies of our repos, to aid development.

Two avenues to consider for such support.

1. "Editable" setups: `nextstrain setup` pointed at local directory (like Python's "editable installs")
2. No setup required: `nextstrain run` pointed at a local directory (like `nextstrain build`)

### "Editable" setups

Accept a local file path/URL when setting up a version.

    nextstrain setup avian-flu@dev=file://$PWD/avian-flu
    nextstrain setup avian-flu@dev=$PWD/avian-flu
    nextstrain setup avian-flu@dev=./avian-flu

    cd avian-flu
    nextstrain setup avian-flu@dev=.

Invoked the same as any other version.

    nextstrain run avian-flu@dev segment-focused /tmp/analysis

Can be approximated currently by:

```console
$ ln -sv ~/nextstrain/measles ~/.nextstrain/pathogens/measles/local="$(echo -n local | base32)"
$ nextstrain run measles@local …
```

### No setup required

Distinguish managed pathogens vs. non-managed pathogens by
absence/presence of a path separator.

    # Managed (current)
    nextstrain run avian-flu segment-focused /tmp/analysis
    nextstrain run avian-flu@v2 segment-focused /tmp/analysis

    # Unmanaged (new)
    nextstrain run ./avian-flu segment-focused /tmp/analyis

    cd avian-flu
    nextstrain run ./ segment-focused /tmp/analysis

and maybe special case `.` too for convenience?

    nextstrain run . segment-focused /tmp/analysis

Note that we'd still keep pathogen and workflow as separate arguments (i.e. rather than combining them the way `nextstrain build` does) so that
workflows are still *names* and can be resolved via pathogen registration
metadata in the future.

### History

I'd long anticipated some sort of "editable install" feature for `nextstrain setup` to support pathogen development, but [several](https://github.com/nextstrain/cli/pull/419#discussion_r1984007486) [review](https://github.com/nextstrain/avian-flu/pull/103#discussion_r1980458646) [comments](https://github.com/nextstrain/avian-flu/pull/103#discussion_r1980499684) spurred me to think about supporting directory paths directly in `nextstrain run`.

tsibley:
@jameshadfield @joverlee521 You're implicated in previous discussion/commentary here and so I'd definitely appreciate any thoughts you have.

[...]

PR #476, opened 2025-09-17
run: Support unmanaged pathogen repositories with local paths

tsibley:
Resolves: <https://github.com/nextstrain/cli/issues/420>

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,15 @@ supported Python version is always bundled with `nextstrain`.
   Batch job.
   ([#460](https://github.com/nextstrain/cli/pull/460))
 
+* `nextstrain run` now supports an alternative invocation where a path to a
+  local directory that is a pathogen repository can be given instead of a
+  pathogen name (and optionally version).  This allows `nextstrain run` to be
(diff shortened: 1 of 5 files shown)


joverlee521 (approved):
This worked as expected for me when I tried it locally with zika. 

It did flag an unexpected behavior for workflows that are not fully `nextstrain run` compatible, specifically ones that define `workdir` in the Snakefile (e.g. [rabies/ingest](https://github.com/nextstrain/rabies/blob/62f6676989937d42ee2fb64e87ad88c4f7cfb774/ingest/Snakefile#L6C1-L6C9)). The outputs were written to the unmanaged pathogen repository instead of the analysis directory. As expected, I see the same behavior with "normally" setup pathogens

```
nextstrain setup rabies@main 
nextstrain run rabies@main ingest /tmp/rabies-ingest
```
The results end up in `~/.nextstrain/pathogens/rabies/main\=NVQWS3Q\=/ingest/`.

Should `nextstrain run` specifically guard against this by specifying `--directory=build_volume`?

PR #478, opened 2025-09-18
Revert "Ignore environment settings in pathogen repo request"

tsibley:
This reverts commit e042a0ff890279bb389e869bbe990fd362da52fa.

We need to maintain "trust_env" support.

Proxy and CA certificate trust store settings provided via environment variables are crucial for outgoing network access in many tightly-controlled environments, like governmental networks.  They are necessary for _all_ requests, not just some requests.  We want places like CDC (a known user) to be able to use `nextstrain setup` and `nextstrain update`.

netrc is less crucial—it's less used—but it is a longstanding way to provide static credentials for remote hosts, and it can be reasonably expected to work.  We _do_ intend to support private pathogen repos, and netrc currently provides a means for that without embedding credentials into the pathogen source URL.

We can address the issue of invalid credentials causing confusing issues with better error handling instead.

Resolves: <https://github.com/nextstrain/cli/issues/468>
Unresolves: <https://github.com/nextstrain/cli/issues/444>
Related-to: <https://github.com/nextstrain/cli/pull/445>

## Checklist

- [x] Checks pass
- [x] Update changelog

Issue #444, opened 2025-06-27
`nextstrain setup <pathogen>` not working


victorlin:
### Scope

Any device with expired credentials for `api.github.com` in `~/.netrc`.

### Description

`nextstrain setup measles` (and `zika`) are broken (use `--force` to reproduce):

```console
$ nextstrain setup measles --force
Setting up measles@main…
Removing existing setup '/Users/vlin/.nextstrain/pathogens/measles/main=NVQWS3Q=' to start fresh…
Error: Failed to download pathogen setup URL:

    https://nextstrain.org/pathogen-repos/measles/versions/e1d1541c3ac44bc73da5872a2721d4edd10f9651
      ⮡ https://api.github.com/repos/nextstrain/measles/zipball/e1d1541c3ac44bc73da5872a2721d4edd10f9651

The server responded with an error:

    HTTPError: 401 Client Error: Unauthorized for url: https://api.github.com/repos/nextstrain/measles/zipball/e1d1541c3ac44bc73da5872a2721d4edd10f9651

The URL may be incorrect (e.g. misspelled) or no longer
accessible.
```

The URL works fine through an unauthenticated browser session and `curl` command. This seems to be an issue with the Python request and is minimally reproducible with:

```py
import requests
requests.get("https://api.github.com/repos/nextstrain/measles/zipball/e1d1541c3ac44bc73da5872a2721d4edd10f9651")
# <Response [401]>
```

### Possible solutions

1. Ensure the Python request is unauthenticated by setting `requests.Session.trust_env = False`. ([ref](https://github.com/nextstrain/cli/issues/444#issuecomment-3020432365))
2. ⛔️ Catch exceptions and if status=401 re-try with `trust_env=False`. ([ref](https://github.com/nextstrain/cli/issues/444#issuecomment-3021025583))
3. ⛔️ Update the nextstrain.org endpoint to [redirect](https://github.com/nextstrain/nextstrain.org/blob/18fd16081fa8a7e51ae2150d82207a13a9cbd736/src/endpoints/pathogenRepos.js#L96) to another URL. I found at least 3 alternatives that work.
    1. `https://codeload.github.com/<org>/<repo>/legacy.zip/<commit>`

        This is actually what the original URL redirects to if it doesn't return `401`:

        ```sh
        curl -o /dev/null -s -D - "https://api.github.com/repos/nextstrain/measles/zipball/e1d1541c3ac44bc73da5872a2721d4edd10f9651"
        # HTTP/2 302
[...]

PR #479, opened 2025-09-22
Hint at invalid credentials when handling 401/403 errors during pathogen download

tsibley:
Since netrc credentials are automatically and silently added, they may be unknown to the user.  Note their presence and suggest solutions in a hint following the error.  We do not automatically retry without credentials because they may be intended by (and known to) the user.

Ideally this sort of error handling is implemented at a higher (e.g. within nextstrain.cli.run()) or lower (e.g. within nextstrain.cli.requests.Session) level in the call stack so that ~all requests benefit from it, not just the single request targeted here. I'm leaving that for future work, though, as a) that's a more extensive change and b) we expect this specific request to be more affected by stale netrc credentials due to the combination of requests for api.github.com and old versions of the gh CLI.

Resolves: <https://github.com/nextstrain/cli/issues/444>
Related-to: <https://github.com/nextstrain/cli/pull/478>

## Checklist

- [x] Checks pass
- [x] Update changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,14 @@ supported Python version is always bundled with `nextstrain`.
   Batch job.
   ([#460](https://github.com/nextstrain/cli/pull/460))
 
+* When `nextstrain setup <pathogen>` and `nextstrain update <pathogen>`
+  encounter an authentication error downloading from pathogen source URLs, the
+  error message now includes a hint, if applicable, noting the presence of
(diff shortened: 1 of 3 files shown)


genehack (approved):

tsibley:
@victorlin I merged this this morning, but I'd still appreciate your 👀 to see if it suitably addresses #444 in your opinion.

PR #521, opened 2026-04-15
Add custom repr to LaxVersion

victorlin:

This is more robust than relying on repr from the base class which may change across versions, and distinguishes the custom class from its base class.

## Related issue(s)

Fixes [test failures](https://github.com/nextstrain/cli/actions/runs/24470284428/job/71508396288) caused by https://github.com/pypa/packaging/commit/d40bec383fd206a94a95c84638065987cc1450db

## Checklist

- [x] Checks pass
- [x] ~Update changelog~ no functional changes

### DIFF
--- a/nextstrain/cli/util.py
+++ b/nextstrain/cli/util.py
@@ -891,6 +891,9 @@ class LaxVersion(Version):
         self.compliant = compliant
         self.original = original if original is not None else version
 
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Thanks for tracking this down!

Issue #522, opened 2026-04-27
Python 3.8 deprecation warning from cryptography


victorlin:
The following warning started appearing a couple days ago, first seen in [scheduled CI run](https://github.com/nextstrain/cli/actions/runs/24937054209):

> cryptography.utils.CryptographyDeprecationWarning: Python 3.8 is no longer supported by the Python core team and support for it is deprecated in cryptography. The next release of cryptography will remove support for Python 3.8.

This is due to https://github.com/pyca/cryptography/commit/595f6de5a008c9c1d5818437fa97a3c88ddd9d5c released in [version 47.0.0](https://pypi.org/project/cryptography/47.0.0/).

It causes pytest to fail because the warning is emitted as an unmatched exception in doctests.

### Short-term solution

- [x] https://github.com/nextstrain/cli/pull/523

### Long-term solution

- [x] https://github.com/nextstrain/cli/pull/524

joverlee521:
I can add the short term warning suppression.

victorlin:
Thanks! I'll open my work in progress as a draft PR and continue efforts there.

PR #523, opened 2026-04-28
Suppress warning from cryptography about deprecated Python 3.8 support

joverlee521:

Short-term solution for <https://github.com/nextstrain/cli/issues/522>. Similar to previous warning suppression added in <https://github.com/nextstrain/cli/commit/3977fee06321a7bc4c879e45dd32ec3cecc50326>

Cryptography 48.0.0, which has not been released yet, will drop support for Python 3.8,¹ so we should also drop support for 3.8 soon.

¹ <https://cryptography.io/en/latest/changelog/#v48-0-0>

## Checklist

- [x] Checks pass
- [x] Update changelog


joverlee521:
Merging since CI passes successfully without deprecation warning.

PR #526, opened 2026-05-05
Don't block releases on docs

victorlin:

Linkcheck can fail during tag-triggered release runs because the new release docs may not exist on RTD yet. The docs build/linkcheck still runs and reports separately, but release no longer waits on that timing-sensitive check.

## Related issue(s)

- Initial [failure](https://github.com/nextstrain/cli/actions/runs/25397235653/attempts/1) of 11.0.0 release
- [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1778008317534129)

## Checklist

- [x] Checks pass
- [x] ~Update changelog~ N/A, dev change

## Previous implementations

- 6ea66c3

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -364,7 +364,6 @@ jobs:
       - test-dist
       - test-standalone
       - lint
-      - doc
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6


doc/conf.py
```diff
@@ -116,6 +117,10 @@
         # linkcheck job, I guess.  In any case, ENOTIME right now.
         #   -trs, 29 May 2025
         'https://docs.nextstrain.org/projects/cli/en/__NEXT__/',
+
+        # Can't reliably check current release links that might not exist yet
+        # (at least when running linkcheck on the tag-triggered CI run).
+        f'https://docs.nextstrain.org/projects/cli/en/{Version(release).public}/',
```
joverlee521: > This ignores the current release links until the next release, which should be fine.

Hmm, does this mean if we add a borked link to the current version, it will break on the next release CI workflow? I wonder if the release job should just not depend on docs in the CI workflow. It's easy to patch up docs after release

victorlin: Good idea, that's cleaner. Updated PR with that approach.

joverlee521: Another thought is to revisit https://github.com/nextstrain/.github/issues/116, but happy with this approach for now.

joverlee521 (approved):

PR #541, opened 2026-07-27
[view] Update for Auspice v3

jameshadfield:
These changes aren't needed for Auspice v3, as the `auspice view` CLI is backwards compatible and thus will continue to work with the now-deprecated `--datasetDir` / `--narrativeDir` arguments.

Once all runtimes are using Auspice v3 (how can we check this?) we can merge this PR & release a new CLI version.

### DIFF
--- a/nextstrain/cli/command/view.py
+++ b/nextstrain/cli/command/view.py
@@ -131,11 +131,11 @@ def register_parser(subparser):
 
     # Register runners; excludes AWS Batch since that makes no sense.
     #
-    # Note that the --datasetDir and --narrativeDir here might be overriden in
-    # run() below.
+    # NOTE: Auspice v3 uses positional args for datasets/narratives
+    # These paths are worked out in run() and added to opts.default_exec_args
     runner.register_runners(
         parser,
-        exec    = ["auspice", "view", "--verbose", "--datasetDir=.", "--narrativeDir=."],
+        exec    = ["auspice", "view", "--verbose"],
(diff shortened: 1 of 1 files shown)


nextstrain/cli/command/view.py
```diff
@@ -131,11 +131,11 @@ def register_parser(subparser):
 
     # Register runners; excludes AWS Batch since that makes no sense.
     #
-    # Note that the --datasetDir and --narrativeDir here might be overriden in
-    # run() below.
+    # NOTE: Auspice v3 uses positional args for datasets/narratives
```
joverlee521, line 134: > Once all runtimes are using Auspice v3 (how can we check this?) 

Haven't looked at the changes here in detail yet, but wanted to note this can be done by updating the [docker compatible_auspice tag](https://github.com/nextstrain/cli/blob/41868cd81b1158f8266af54d98a78541a8c9f644/nextstrain/cli/runner/docker.py#L110) and probably add a new entry to [conda ENV_FEATURE](https://github.com/nextstrain/cli/blob/41868cd81b1158f8266af54d98a78541a8c9f644/nextstrain/cli/runner/conda.py#L177-L179). There's nothing in place for versioning in the ambient runtime, but maybe we could add something to [test_setup](https://github.com/nextstrain/cli/blob/41868cd81b1158f8266af54d98a78541a8c9f644/nextstrain/cli/runner/ambient.py#L71-L87).

